PR #3, opened 2015-03-17
Added some options.

calculuswhiz:
Mostly just added these for the '-nostdlib' option because it's useful when I write programs in x86.


danielshahaf:
Merged in e3188ff0e80bd9e4515fc50ce8c2e78b3f2f4fdc.

Sorry for the delay: zsh doesn't use github. If you send patches to zsh-workers_at_zsh.org they'll probably be applied much more quickly.

Thanks for the patch!

PR #4, opened 2015-05-13
Completion: git commit object improvements

blueyed:
Submitted to zsh-workers: http://www.zsh.org/mla/workers/2015/msg01023.html


blueyed:
This has been committed to Zsh.

PR #5, opened 2015-06-20
Correct typo in compinstall

yousefamar:
Diff says it all.


danielshahaf:
Merged in ac5d83bf56503d79de3d60a236326d687a491ecd.

Sorry for the delay: zsh doesn't use github. If you send patches to zsh-workers_at_zsh.org they'll probably be applied much more quickly.

Thanks for the patch!

PR #6, opened 2015-10-10
Fixed example (`$color`)

dritter:
The correct color variable is `$color`, not `$colors`.


danielshahaf:
I've merged this in commit dd2aa858073e57a52c541cce04fb99ba7553c6b8.

Sorry for the delay: zsh doesn't use github.  If you send patches to zsh-workers_at_zsh.org they'll probably be applied much more quickly.

Thanks for the patch!

dritter:
Thanks! :)

PR #8, opened 2016-05-12
Add git-subtree completion

forivall:
I had this sitting in a closed pr in zsh-completions for a long time (https://github.com/zsh-users/zsh-completions/pull/372). This could also definitely use some polish.

### DIFF
--- a/Completion/Unix/Command/_git
+++ b/Completion/Unix/Command/_git
@@ -1747,6 +1747,92 @@ _git-submodule () {
   return ret
 }
 
+(( $+functions[_git-subtree] )) ||
+_git-subtree () {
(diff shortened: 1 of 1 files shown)


nicoulaj:
Hi, this is just a mirror, you should post your patch to zsh-workers mailing list instead.

forivall:
http://www.zsh.org/mla/users//2016/msg00418.html

I just haven't had the time to finish it up.

okapia:
Just thought I'd give you a prod. Would be nice to have it finished so that it can be checked in.

http://www.zsh.org/mla/users//2016/msg00427.html

PR #10, opened 2016-10-23
yum completion

pseyfert:
adding the following yum commands:
- `update-to` (just what `update` expands)
- `upgrade-to` (just what `update` expands)
- `history` (the actual tricky one)
- `help` (printing the available commands)
- `load-transaction` (`_files`)
- `load-ts` (`_files`)
- `check` (just what the man page suggested)
- `reinstall` (just what `erase` does - suggest installed packages)
- `downgrade` (just what `erase` does - suggest installed packages)
- `repolist` (just what the man page suggested)
- `distribution-synchronization` (just what `erase` does - suggest installed packages)
- `distro-sync` (just what `erase` does - suggest installed packages)

`history` still needs work. For some of the commands one should supply a transaction id (i.e. just a number from 1 to `yum history stats | grep Transactions | sed "s/.*: //"` or `last` or `all`) or a range (e.g. `5..23`). At the moment I just have `all` `last` as well as transaction IDs 1-3 hard coded for testing.

Also, I didn't implement all `history` commands - I left out those where I couldn't easily test if I would suggest a valid command, such as in `yum history rollback`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,8 @@
 2016-10-20  Peter Stephenson  <p.stephenson@samsung.com>
 
+	* 39688: Src/pattern.c: Alternative fix: next pointer is
+	irrelevant anyway as there's only one P_EXACTLY in a P_PURES.
+
 	* 39683: Src/pattern.c: Update scan pointer in pattern after
 	possible reallocation for meta handling.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2016-10-20  Peter Stephenson  <p.stephenson@samsung.com>
+
+	* 39683: Src/pattern.c: Update scan pointer in pattern after
+	possible reallocation for meta handling.
+
 2016-10-19  Barton E. Schaefer  <schaefer@zsh.org>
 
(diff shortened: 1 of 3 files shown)


danielshahaf:
> adding the following yum commands:

Please send this to zsh-workers@zsh.org; the zsh project doesn't use
github.  Thanks!

P.S. The first hunk of your patch has an indentation error (four spaces
instead of two tabs).

pseyfert:
okay. As a patch file attached to the email?

Looking back at the master I see the formatting is before my patch mixed (spaces vs tabs) already. What's the desired indentation style?

danielshahaf:
> okay. As a patch file attached to the email?

Yes, attachments are fine.

> Looking back at the master I see the formatting is before my patch mixed (spaces vs tabs) already. What's the desired indentation style?

Whatever is consistent with the rest of the file.  I only noticed this
because the patch added a line using spaces inside a block ($_yum_cmds)
that uses tabs, so the diff looked funny.

Thanks!

pseyfert:
closed by 2ba985fe5b8d3c61aa8c066c9cb14165a409e90b

PR #12, opened 2017-01-07
-Dproperty=/path_complete

dezza:
I think this change is desirable since line 86 already proposes a _default completion and path completion if magic_equal_subst is enabled and it is never reached if -D is not matched on -D* e.g. "-Dproperty" as most people use -D properties as I'm aware (Not a Java programmer).


dezza:
That makes sense. I'm just starting to get to know this syntax so bear with me :)

You can submit it if you want if not I will.

okapia:
merged in b8b5f18f3fcf

PR #13, opened 2017-02-19
Fix VCS_INFO outputing messages to stderr

nicoulaj:
VCS_INFO invokes `git diff --no-ext-diff --ignore-submodules=dirty --quiet --exit-code`, which can output messages to stderr (for instance if a committed file cannot be read by the current user).


nicoulaj:
Posted on zsh-workers/40586
http://www.zsh.org/mla/workers/2017/msg00322.html

PR #16, opened 2017-03-26
Fix seg fault on user autocomplete under msys

cyberwasp:
Under windows (MSYS2 or Cygwin) user name can contains local characters. So it need metafy.


danielshahaf:
> Under windows (MSYS2 or Cygwin) user name can contains local characters. So it need metafy.

Thanks for the PR.

zsh does not use github.  Could you please send the patch to zsh-workers@zsh.org?

cyberwasp:
Sent. Thanks for the tip.

PR #17, opened 2017-04-03
Completion: add support for the "Include" command _ssh_hosts

DanielWeigl:
OpenSSH has added an "Include" command for the ssh_config file, to reference additional configuration files, see http://man.openbsd.org/ssh_config#Include (since version 7.3)

Add support for hostname detection to _ssh_hosts

### DIFF
--- a/Completion/Unix/Command/_ssh
+++ b/Completion/Unix/Command/_ssh
@@ -664,7 +664,9 @@ _ssh_users () {
 
 _ssh_hosts () {
   local -a config_hosts
-  local config
+  local -a config
+  local config_file key args host cfgfile
+  integer cnt
   integer ind
 
   # If users-hosts matches, we shouldn't complete anything else.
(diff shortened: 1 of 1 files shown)


danielshahaf:
> OpenSsh has added an "Include" command for the ssh_config file, to
> reference additional configuration files, see
> http://man.openbsd.org/ssh_config#Include (since version 7.3)

Thanks for the PR + docs reference.

zsh does not use github.  Moreover, two other patches for this have been
sent in the "patch for ssh completion" thread on workers@ this month.

Could you please send the patch to workers@?  Ideally also with
comparison against the other competing patches.  Thanks again.

> Add support for hostname detection to _ssh_hosts

DanielWeigl:
Okay, thank you for the info, did not realize the workers@ list. just noted that here on github were merged pull requests too and did not further check them.

Ill close this PR then

PR #18, opened 2017-11-10
Specify system python in dnf completion

malramsay64:
The python executable in the python_exec variable was the python[2|3]
executable found in $PATH. When the python[2|3] executable was not the
system installed version the completion command raised an `ImportError`
as the 'dnf' module was not found.

This commit just completely specifies the python version to be the
system version `/usr/bin/python[2|3]`, allowing completion within a
python dev environment.


okapia:
Merged.

The function already has /usr/bin/dnf hardcoded so hardcoding /usr/bin/python is no worse than it already is. If it proves problematic, we could have it search for a python in the same location as dnf first.

PR #21, opened 2017-12-31
updated nmcli completions to be compatible with recent version 1.8.4

mmakowski:
The existing nmcli completions are outdated after significant changes to the command line interface post 1.0.0. This patch provides fairly comprehensive completions for a recent version of the tool.

Addresses [zsh-completions issue #434](https://github.com/zsh-users/zsh-completions/issues/434).


Completion/Linux/Command/_nmcli
```diff
@@ -1,244 +1,369 @@
-#compdef nmcli
+#compdef _nmcli nmcli
```
okapia: Is that correct? Is there an _nmcli command? Or are you mixing up the syntax of #compdef directives with the compdef command?

Arguably the function should be named _networkmanager or similar rather than _nmcli. Given the complete change of function we should perhaps consider renaming.

Completion/Linux/Command/_nmcli
```diff
@@ -1,244 +1,369 @@
...
-typeset -A opt_args
+# compatible with nmcli 1.8.4
 
-if [[ -z $_nmcli_version ]]; then
-    _nmcli_version="${"$(_call_program nmcli $words[1] --version)"##*version }"
-fi
+_nmcli() {
+    local line    _arguments -C \
+        {'(-terse)-t','(-t)-terse'}"[terse output]" \
+        {'(-pretty)-p','(-p)-pretty'}"[pretty output]" \
```
okapia: Could you perhaps reindent this according the the normal zsh conventions which is just two spaces of indentation. See Etc/completion-style-guide. I understand that the old completion function for nmcli used 4 but as nothing from that has been preserved, there's no need to preserve it to preserve history,

Completion/Linux/Command/_nmcli
```diff
@@ -1,244 +1,369 @@
...
+        {'(-mode)-m','(-m)-mode'}"[output mode]: :(tabular multiline)" \
+        {'(-colors)-c','(-c)-colors'}"[whether to use colors in output]: :(auto yes no)" \
+        {'(-fields)-f','(-f)-fields'}"[specify fields to output]:fields:_nmcli_fields" \
+        {'(-get-values)-g','(-g)-get-values'}"[shortcut for -m tabular -t -f]:fields:_nmcli_fields" \
+        {'(-escape)-e','(-e)-escape'}"[escape columns separators in values]: :(yes no)" \
+        {'(-ask)-a','(-a)-ask'}"[ask for missing parameters]" \
+        {'(-show-secrets)-s','(-s)-show-secrets'}"[allow displaying passwords]" \
+        {'(-wait)-w','(-w)-wait'}'[set timeout waiting for finishing operations]:timeout (s)' \
```
okapia: A minor point but I'd spell out "seconds" instead of just "s"

I don't know if this is a direct copy from nmcli documentation but the English is a bit dodgy in both of these last two descriptions. Perhaps: "allow passwords to be displayed" or "allow display of passwords"
And: "set time limit on wait for operations to finish" or maybe just "set timeout for operations"

Completion/Linux/Command/_nmcli
```diff
@@ -1,244 +1,369 @@
...
+        {'(-version)-v','(-v)-version'}"[show program version]" \
+        {'(-help)-h','(-h)-help'}"[print help]" \
```
okapia: It is quite rare that -version and -help options can be combined with other options so you may just be able to use something like:
  '(- *)-h'{,elp}'[display usage information]'
The (- *) excludes all other options and the args.

Completion/Linux/Command/_nmcli
```diff
@@ -1,244 +1,369 @@
...
+        c*) _nmcli_connection ;;
+        d*) _nmcli_device ;;
+        a*) _nmcli_agent ;;
+    esac
 }
 
-(( $+functions[_nmcli_con_ids] )) || _nmcli_con_ids() {
-    _nmcli_is_running || return
+_nmcli_fields() {
+    _values -s , 'fields to be printed' \
```
okapia: By convention, description should be singular, i.e. "field". Only one is completed at a time even if a list is allowed.

Completion/Linux/Command/_nmcli
```diff
@@ -1,244 +1,369 @@
...
-    _nmcli_version="${"$(_call_program nmcli $words[1] --version)"##*version }"
-fi
+_nmcli() {
+    local line    _arguments -C \
+        {'(-terse)-t','(-t)-terse'}"[terse output]" \
+        {'(-pretty)-p','(-p)-pretty'}"[pretty output]" \
+        {'(-mode)-m','(-m)-mode'}"[output mode]: :(tabular multiline)" \
+        {'(-colors)-c','(-c)-colors'}"[whether to use colors in output]: :(auto yes no)" \
+        {'(-fields)-f','(-f)-fields'}"[specify fields to output]:fields:_nmcli_fields" \
+        {'(-get-values)-g','(-g)-get-values'}"[shortcut for -m tabular -t -f]:fields:_nmcli_fields" \
```
okapia: By convention, description should be singular, i.e "field". Only one is completed at a time, even if there can be a list of them.

Completion/Linux/Command/_nmcli
```diff
@@ -1,244 +1,369 @@
...
 
-if [[ -z $_nmcli_version ]]; then
-    _nmcli_version="${"$(_call_program nmcli $words[1] --version)"##*version }"
-fi
+_nmcli() {
+    local line    _arguments -C \
+        {'(-terse)-t','(-t)-terse'}"[terse output]" \
+        {'(-pretty)-p','(-p)-pretty'}"[pretty output]" \
+        {'(-mode)-m','(-m)-mode'}"[output mode]: :(tabular multiline)" \
+        {'(-colors)-c','(-c)-colors'}"[whether to use colors in output]: :(auto yes no)" \
```
okapia: Per-match descriptions are missing. The first should be "mode", the second could be just "value" if you can't think of anything better.

Completion/Linux/Command/_nmcli
```diff
@@ -1,244 +1,369 @@
...
 
-local context state line expl
-typeset -A opt_args
+# compatible with nmcli 1.8.4
 
-if [[ -z $_nmcli_version ]]; then
-    _nmcli_version="${"$(_call_program nmcli $words[1] --version)"##*version }"
-fi
+_nmcli() {
+    local line    _arguments -C \
```
okapia: Have two lines become wrapped here? When using _arguments' -C option, you should also declare:
  local curcontext="$curcontext" state line

Completion/Linux/Command/_nmcli
mmakowski: Indeed, the `_nmcli` is unnecessary -- this is my first attempt to write zsh completions, and I based it partly on [a blog post](http://mads-hartmann.com/2017/08/06/writing-zsh-completion-scripts.html) that contains this pattern.

With regards to the rename, it seemed to me that it makes sense to stick to the convention where the completion function's name corresponds to the command name -- as far as I understand there is no point having both `_nmcli` and `_networkmanager`, the former for old version of `nmcli` and the latter for new -- but please let me know if you would prefer that it is renamed, and I will do that.

Completion/Linux/Command/_nmcli
mmakowski: Yes, that must have been an accidental change prior to the commit.

With regards to `local curcontext="$curcontext" state line`, should I still declare `curcontext` and `state` if I don't use them in this function?

mmakowski:
Thank you for the feedback! I will make changes as suggested.

Completion/Linux/Command/_nmcli
okapia: There's no point keeping the old _nmcli at all, I wasn't suggesting that.
The usual convention is actually closer to that used for deb/rpm package naming. So, for example, subversion completion is in _subversion rather than _svn. With that in mind, I'd prefer it be renamed.

Completion/Linux/Command/_nmcli
okapia: Yes. When you're using _arguments' state (:->args) syntax, you are asking _arguments to modify those variables. The value of $curcontext is being used indirectly in later calls _description when looking up styles. Note that curcontext needs a coy of the existing value.

mmakowski: Thanks for the explanation!

Completion/Linux/Command/_nmcli
mmakowski: That makes sense, I will rename the function.

okapia:
With a few minor tweaks, I have now merged this. Thanks for the contribution.

One thing I'm a little unsure of is the use of _describe in a few places. _describe is meant to be used where there are per-match descriptions. If you've just got single values, use _wanted and compadd. The line may be longer but it goes through vastly less code internally. I don't really use NetworkManager so it is tricky to confirm. But if you make a further update you can do a fresh PR.

Thanks again.

mmakowski:
Thank you for the review and the merge! Regarding `_describe`, I only used it because it was a pattern I have seen in the existing implementation of `_nmcli`. I will be happy to change it to `_wanted`/`compadd` if I get round to making any more changes.

PR #23, opened 2018-02-25
Add kdeconnect-cli autocompletion

mtijink:
Adds autocompletion for kdeconnect-cli (part of KDE Connect).

Feedback is welcome, as this is my first zsh patch.

### DIFF
--- a/Completion/X/Command/_kdeconnect-cli
+++ b/Completion/X/Command/_kdeconnect-cli
@@ -1,0 +1,33 @@
+#compdef kdeconnect
+
+_kdeconnect-cli-devices-id() {
+  compadd $(kdeconnect-cli --list-available --id-only)
+}
+
+_arguments \
+  '(-l --list-devices -a --list-available)'{-l,--list-devices}'[list all devices]' \
+  '(-l --list-devices -a --list-available)'{-a,--list-available}'[list available (paired and reachable) devices]' \
+  '--id-only[make --list-devices or --list-available print only the devices id, to ease scripting]' \
+  '--refresh[search for devices in the network and re-establish connections]' \
+  '(--pair --unpair)--pair[request pairing with the specified device]' \
+  '--ring[find the device by ringing it.]' \
+  '(--pair --unpair)--unpair[stop pairing to the specified device]' \
+  '(--ping --ping-msg)--ping[send a ping to the device]' \
+  '(--ping --ping-msg)--ping-msg[send a ping to the device with the specified message]:message' \
+  '--share[share a file to the device]:file:_files' \
+  '--list-notifications[display the notifications on the device]' \
+  '--lock[lock the specified device]' \
+  '--send-sms[send an SMS. Requires --destination]:message' \
+  '--destination[phone number to send the SMS to]:phone number' \
+  '(-d --device -n --name)'{-d,--device}'[device ID]:id:_kdeconnect-cli-devices-id' \
+  '(-d --device -n --name)'{-n,--name}'[device name]:name' \
+  '--encryption-info[get encryption info about the device]' \
+  '--list-commands[list remote commands and their ids]' \
+  '--execute-command[execute a remote command]:command id' \
+  '(-k --send-keys)'{-k,--send-keys}'[send keys to the specified device]' \
+  '--my-id[display this device''s id and exit]' \
+  '(-)'{-h,--help}'[display help and exit]' \
+  '(-)'{-v,--version}'[display version information and exit]' \
+  '(-)--author[show author information and exit]' \
+  '(-)--license[show license information and exit]' \
+  '--desktopfile[the base file name of the desktop entry for this application]:file'


Completion/X/Command/_kdeconnect-cli
```diff
@@ -0,0 +1,35 @@
+#compdef kdeconnect-cli
+
+_kdeconnect-cli() {
+  _kdeconnect-cli-devices-id() {
```
okapia: For Zsh autoloaded functions, you don't need the _kdeconnect-cli() { } wrapping the entire contents of the file. That's superfluous and should be removed.

Completion/X/Command/_kdeconnect-cli
```diff
@@ -0,0 +1,35 @@
+#compdef kdeconnect-cli
+
```
okapia: The normal naming convention for zsh completion functions is to use the name of the software rather than the command name. So in this case, it should probably be just _kdeconnect
Similarly, we have _subversion for svn (and svn-admin), _perforce for p4, _ansible for ansible-playbook and so on.

Completion/X/Command/_kdeconnect-cli
```diff
@@ -0,0 +1,35 @@
...
+_kdeconnect-cli() {
+  _kdeconnect-cli-devices-id() {
+    compadd $(kdeconnect-cli --list-available --id-only)
+  }
+
+  _arguments \
+    '(-l --list-devices)'{-l,--list-devices}'[List all devices]' \
+    '(-a --list-available)'{-a,--list-available}'[List available (paired and reachable) devices]' \
```
okapia: Normal convention for descriptions is not to capitalize the first work (and not to use a fullstop). So: 
`[list all devices]`

Completion/X/Command/_kdeconnect-cli
```diff
@@ -0,0 +1,35 @@
...
+    '--id-only[Make --list-devices or --list-available print only the devices id, to ease scripting]' \
+    '--refresh[Search for devices in the network and re-establish connections]' \
+    '--pair[Request pairing to a said device]' \
```
okapia: Which device is the "said" device? When has it been said? Perhaps this has been copied from kdeconnect-cli's help output but I think the wording would be clearer without the word "said". Especially for non-native English speakers.

Completion/X/Command/_kdeconnect-cli
```diff
@@ -0,0 +1,35 @@
...
+    '--ring[Find the said device by ringing it.]' \
+    '--unpair[Stop pairing to a said device]' \
+    '--ping[Sends a ping to said device]' \
+    '--ping-msg:Same as ping but you can set the message to display:' \
```
okapia: This can be constructed more nicely. For example:
`'--ping-msg[send a ping to a device with a custom message to display]:message' \`

Does it allow `--ping-msg=message` or only `--ping-msg message`?

Completion/X/Command/_kdeconnect-cli
```diff
@@ -0,0 +1,35 @@
...
+    '--share:Share a file to a said device:_files' \
+    '--list-notifications[Display the notifications on a said device]' \
+    '--lock[Lock the specified device]' \
+    '--send-sms:Sends an SMS. Requires destination:' \
```
okapia: Like `--ping-msg`, you should use a description of the option in square brackets and a description of the argument after the colon. e.g:
`'--send-sms[send an SMS]:destination'`

or is the destination supposed to be specified with --destination?

Starting descriptions with the command form of the verb is preferred over simple present tense, "send" instead of "sends" in this case. It is also good to be consistent - you have "display" and not "displays" earlier.

Completion/X/Command/_kdeconnect-cli
```diff
@@ -0,0 +1,35 @@
...
+    '--destination:Phone number to send the message:' \
+    '(-d --device)'{-d,--device}':Device ID:_kdeconnect-cli-devices-id' \
+    '(-n --name)'{-n,--name}':Device Name:' \
+    '--encryption-info[Get encryption info about said device]' \
+    '--list-commands[Lists remote commands and their ids]' \
+    '--execute-command:Executes a remote command by id:' \
+    '(-k --send-keys)'{-k,--send-keys}'[Sends keys to a said device]' \
+    '--my-id[Display this device''s id and exit]' \
+    '(-h --help)'{-h,--help}'[Displays help.]' \
```
okapia: --help style options often exclude all other options from following. So this perhaps might be `'(-)'` instead of `'(-h --help)'`. Same goes for other options like --version, --author and --license. I suspect there are also further cases of options that are mutually exclusive. Can you specify both a device name and ID?

mtijink:
Thanks for the feedback, I incorporated the changes.

okapia:
Thanks. Merged in f5d0111

I made a couple of very minor tweaks–in particular we can't rely on the `rc_quotes` option in completion functions.

PR #24, opened 2018-04-05
Add completion script for Visual Studio Code

Mellbourn:
This script adds command line completion for the `code` command, which is used to start Visual Studio Code (https://code.visualstudio.com/). Code is ranked the most popular developer environment in the [Stack Overflow Developer Survey 2018](https://insights.stackoverflow.com/survey/2018/#development-environments-and-tools), so I think there is a strong case for including completion functionality directly in zsh.

This implementation gives completion for all documented parameters and has specialized completion for parameters that don't take filenames as parameters, for example `--add` only takes directories as arguments.
I have tested that it works well on macOS and Windows Subsystem for Linux (Ubuntu).

I suggested this script to the maintainers of Visual Studio Code, but they felt it was out of scope of the product itself, and suggested I provide the script externally (Microsoft/vscode#47101). So it would be great if it could be added directly to zsh instead.

I have also [submitted this completion script to the zsh-completions repository](https://github.com/zsh-users/zsh-completions/pull/561), but it was pointed out to me that it could be a better idea to have the completion directly in `zsh`. So if you accept this pull request, I can withdraw the pull request from zsh-completions, since it will be redundant there.


danielshahaf:
In --level and --locale, I suggest to move the example/default values to after the first colon.

In the helper function, `${${(f)"$(locale -a)"}%%.*}` should be faster and split on newlines only (per POSIX). I think it would fit better as Type/_locales, though—doesn't that exist already?

Excuse brevity (touch keyboard).

Completion/Unix/Command/_code
```diff
@@ -0,0 +1,42 @@
...
+  locales=($(locale -a | sed -e 's/\..*$//'))
+  _describe 'locale' locales
+}
+
+arguments=(
+  {-d,--diff}'[compare two files with each other]:file to compare:_files:file to compare with:_files'
+  {-a,--add}'[add folder(s) to the last active window]:folder:_directories'
+  {-g,--goto}'[open a file at the path on the specified]'
+  {-n,--new-window}'[force to open a new window]'
+  {-r,--reuse-window}'[force to open a file or folder in the last active window]'
```
okapia: Are these last two options mutually exclusive. If so, we can indicate to zsh that only one should be used. It is also usually the case that long and short option forms are mutually exclusive of each other. Furthermore, the English in these descriptions is not great. I know it comes from code's --help output but the word "force" when used as a verb like this should have an object. So, I'd suggest something like:

    '(-n --new-window -r --reuse-window)'{-n,--new-window}'[open a new window]'

And similarly for -r.

Completion/Unix/Command/_code
```diff
@@ -0,0 +1,42 @@
...
+  {-v,--version}'[print version]'
+  {-h,--help}'[print usage]'
+  '--extensions-dir[set the root path for extensions]:root path:_directories'
+  '--list-extensions[list the installed extensions]'
+  '--show-versions[show versions of installed extensions, when using --list-extension]'
+  '--install-extension[installs an extension]:id or path:_files -g "*.vsix"'
+  '--uninstall-extension[uninstalls an extension]:id or path:_files -g "*.vsix"'
+  '--enable-proposed-api[enables proposed api features for an extension]:extension id:'
+  '--verbose[print verbose output (implies --wait)]'
+  '--log[log level to use. Default is info]:level:(critical error warn info debug trace off)'
```
okapia: Usual zsh completion way to indicate defaults is with square brackets in the argument description. That ensures it is on screen when you need it, e.g:

    '--log[specify log level to use]:level [info]:(critical error warn info debug trace off)'

I also add the word "specify" to the start of the description which is very common. In most cases, the first word of per-match descriptions will be a verb.

Completion/Unix/Command/_code
```diff
@@ -0,0 +1,42 @@
...
+
+arguments=(
+  {-d,--diff}'[compare two files with each other]:file to compare:_files:file to compare with:_files'
+  {-a,--add}'[add folder(s) to the last active window]:folder:_directories'
+  {-g,--goto}'[open a file at the path on the specified]'
+  {-n,--new-window}'[force to open a new window]'
+  {-r,--reuse-window}'[force to open a file or folder in the last active window]'
+  {-w,--wait}'[wait for the files to be closed before]'
+  '--locale[the locale to use (e.g. en-US or zh-TW)]:locale:__code_locales'
+  '--user-data-dir[specifies the directory that user data is in]:directory:_directories'
```
okapia: I'd change the verb "specifies" to "specify". Similarly later, I'd change enables to enable, uploads to upload and (un)installs to (un)install. It is more consistent to use the same form for the verbs in descriptions and you have "print", "set" and "list" elsewhere.  Usual zsh convention is this short verb form.

Completion/Unix/Command/_code
```diff
@@ -0,0 +1,42 @@
...
+  '--log[log level to use. Default is info]:level:(critical error warn info debug trace off)'
+  {-s,--status}'[print process usage and diagnostics information]'
+  {-p,--performance}'[start with the "Developer: Startup Performance" command enabled]'
+  '--prof-startup[run CPU profiler during startup]'
+  '--disable-extensions[disable all installed extensions]'
+  '--inspect-extensions[allow debugging and profiling of extensions]'
+  '--inspect-brk-extensions[allow debugging and profiling of extensions with the extension host being paused after start]'
+  '--disable-gpu[disable GPU hardware acceleration]'
+  '--upload-logs[uploads logs from current session to a secure endpoint]'
+  '--max-memory[max memory size for a window]:size (in Mbytes):'
```
okapia: units in parentheses is a common convention anyway, the word "in" isn't needed.
Perhaps start the other description with "specify ", though.

Completion/Unix/Command/_code
```diff
@@ -0,0 +1,42 @@
...
+  '*:filename:_files'
+)
+
+_arguments $arguments
```
okapia: We can pass -s and -S to _arguments - short forms like -wn for -w -n are allowed and no options are allowed after --. So this can be:
_arguments -s -S $arguments

Completion/Unix/Command/_code
```diff
@@ -0,0 +1,42 @@
+#compdef code
+
+local arguments
+
+__code_locales() {
+  local -a locales
+  locales=($(locale -a | sed -e 's/\..*$//'))
+  _describe 'locale' locales
+}
```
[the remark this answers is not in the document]
Mellbourn: Agreed. I hard coded the list of actual possibilities. That makes the list straightforward, and more accurate.

Completion/Unix/Command/_code
```diff
@@ -0,0 +1,42 @@
...
+
+arguments=(
+  {-d,--diff}'[compare two files with each other]:file to compare:_files:file to compare with:_files'
+  {-a,--add}'[add folder(s) to the last active window]:folder:_directories'
```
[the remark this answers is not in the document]
Mellbourn: Good point. I have added the asterisk. I changed to directory/directories. Wasn’t totally sure about how to handle the “(ies)” part. I chose “directory (or directories)”. Feel free to suggest a better wording.

Completion/Unix/Command/_code
Mellbourn: Good point. I followed your advice regarding -r and -n. I have also made all long and short arguments exclusive

Completion/Unix/Command/_code
Mellbourn: Ok, changed the defaults to be in square brackets. Added "specify"

Completion/Unix/Command/_code
Mellbourn: Changed lots of verb cases and used “specify” more

Completion/Unix/Command/_code
Mellbourn: Removed “in”

Completion/Unix/Command/_code
Mellbourn: Added parameters. I had no idea about the `--` trick.

Thank you for all your comments!
I made some further improvements: 
* `--upload-logs` can now complete the confirmation.
* Corrected `--wait` description
* Made it easier to end `--goto` parameter with a colon
* Added an optional `=` to some options that can accept it (unclear to me exactly which do, but it is not all)

Please let me know if I got something wrong or there is anything else I should improve.

Completion/Unix/Command/_code
```diff
@@ -0,0 +1,36 @@
+#compdef code
+
+local arguments
+
+arguments=(
+  '(-d --diff)'{-d,--diff}'[compare two files with each other]:file to compare:_files:file to compare with:_files'
+  '(-a --add)'\*{-a,--add}'[add specified directory (or directories) to the last active window.]:directory:_directories'
```
okapia: You don't want the (-a --add) exclusion as well as \*.
The * prevents -a excluding -a (i.e. itself) which it does by default. The option can be repeated so you can mix -a and --add. So no exclusion is needed.

I wouldn't bother with the "(or directories)". The fact that the option can be repeated conveys the information that more than one directory can be added. Admittedly, _arguments doesn't make that as obvious as it might in completion listings but it'd be better to try to make it do so for all completions than to use convoluted wording in each one.

Completion/Unix/Command/_code
```diff
@@ -0,0 +1,36 @@
...
+  '(-g --goto)'{-g,--goto}'[open a file at the path on the specified line and character position]:file\:line[\:character]:_files -r \:'
```
okapia: Unlike earlier, I don't have VS code available but I wouldn't be surprised if the -g option can also be repeated. Could be worth checking.

Completion/Unix/Command/_code
```diff
@@ -0,0 +1,36 @@
...
+  '(-n --new-window -r --reuse-window)'{-n,--new-window}'[open a new window]'
+  '(-n --new-window -r --reuse-window)'{-r,--reuse-window}'[open a file or directory in the last active window]'
+  '(-w --wait)'{-w,--wait}'[wait for the files to be closed before returning]'
+  '--locale=[specify the locale to use]:locale (e.g. en-US or zh-TW):(de en en-US es fr it ja ko ru zh-CN zh-TW bg hu pt-br tr)'
+  '--user-data-dir[specify the directory that user data is in]:directory:_directories'
+  '(-v --version)'{-v,--version}'[print version]'
+  '(-h --help)'{-h,--help}'[print usage]'
```
okapia: I can't verify with code at the moment but it is common for options like -h and -V to not be applicable in combination with any other options. In which case the exclusion list can be just (- *)

Completion/Unix/Command/_code
```diff
@@ -0,0 +1,36 @@
...
+  '--extensions-dir[specify the root path for extensions]:root path:_directories'
+  '--list-extensions[list the installed extensions]'
+  '--show-versions[show versions of installed extensions, when using --list-extension]'
+  '--install-extension[specify extension to install]:id or path:_files -g "*.vsix"'
+  '--uninstall-extension[specify extension to uninstall]:id or path:_files -g "*.vsix"'
```
okapia: This is a very pedantic point but in completion, it is common to use the glob qualifier (-.) with _files -g.
So _files -g "*.vsix(-.)"
That only matters if you have something like a directory named with the extension.

Completion/Unix/Command/_code
Mellbourn: Agreed. Removed `(-a --add)` exclusion and redundant pluralization.

Completion/Unix/Command/_code
Mellbourn: Nope. When I test, it seems to ignore the second `-g`

Completion/Unix/Command/_code
Mellbourn: Set exclusion list to `(- *)`

Maybe the obverse should be done too:
If any parameters have been added, disallow `-v` and `-h`?

Completion/Unix/Command/_code
Mellbourn: Done.

Thanks again for your comments!

okapia:
Merged in 5e10acc

Thanks for the contribution and for your patience with the petty comments.

PR #26, opened 2018-08-07
__git_files: fix double quotation of star

mcornella:
This fixes the double quote introduced by aa160fc8, so that the end result of the parameter expansion is `$pref` followed by `\*` (a quoted star), meaning `git ls-files` gets a literal star, not an expanded list of matching files, as was intended in commit cc7437bf.

This fixes the use case of trying to `git-add` files from the parent directory (`git add ../<TAB>`). I haven't tested other use cases.


danielshahaf:
Merged in 99fd8c0fe97b.  Thanks for the patch!

(and thanks @okapia for the ping)

mcornella:
Yay thanks!

PR #29, opened 2018-10-03
Add completion support for Gits --color-moved option and color settings

mineo:
--color-moved was added in Git 2.15.


okapia:
Merged in 11c84dce

Thanks for the contribution.
It seems the output from `git diff -h` isn't especially helpful so these weren't picked up in the normal update checks.

PR #30, opened 2018-10-17
change history for fc, fix event not found error

soyuka:


okapia:
When I do `which history`, I get `history: shell built-in command`. So it looks like it is an alias rather than a function.

Fortunately, we can prevent aliases expanding inside functions. Where you load zcalc in your `.zshrc`, use `autoload -U zcalc` where the `-U` tells it to disable alias expansion. This is good practice for all autoloadable functions.

I don't as such object to having `fc` in the code but I prefer to understand a problem rather than just paper over it.

soyuka:
> I don't as such object to having fc in the code but I prefer to understand a problem rather than just paper over it.

I understand, thanks for studying the use case, I guess I'm going to close this patch then thanks!

okapia:
> @okapia I guess the `1` in the alias' expansion breaks things?

Yes. I'm somewhat inclined to apply the patch anyway, however. Just because people are more likely to have aliases (or functions) named `history` than `fc`. What do you think?

soyuka:
> Yes. I'm somewhat inclined to apply the patch anyway, however. Just because people are more likely to have aliases (or functions) named history than fc. What do you think?

Because fc is the replacement to history I think that the patch is indeed safe.

soyuka:
> Yes, the patch is safe. I wouldn't call fc the replacement to history, however. fc is older, dating at least to the 1988 release of ksh.

#til thanks!

> @soyuka: did you check your .zshrc? Were you able to fix the issue for yourself with autoload -U?

I'd need to check on another computer but your solution seems obvious. Thanks, I'll re-open if the issue isn't fixed.

danielshahaf:
@okapia The need for wrappers not to break the wrappee's CLI API is ubiquitous (for example, compsys relies on it), so ultimately this is a zshrc bug.

I would argue to fix it anyway in the name of usability, but using `builtin history` would break `history ()` wrappers that _don't_ break the builtin's API, and using `fc` because it's unlikely to be incompatibly shadowed… I'm not _against_ it but it feels like whack-a-mole. (What would we do when someone incompatibly shadows `fc`? When another function calls `history`?)

Can we solve _all_  variations of this bug? E.g., by forcing -U for stdlib fpath entries, or by warning when a shadowing alias is created?

Continue the discussion on -workers@?

PR #32, opened 2019-01-28
Fix a typo in the git tag completion.

333fred:
Noticed a small typo in the completion list for `git tag`. Looking at the contributing section on the Git page, I believe that this appropriate to submit via pull request. If it is not, let me know and I can submit it via a mailing list.


danielshahaf:
Thanks for that! Merged in 4f6ac08af3ffd9ce13912a50cbeed6546a2440e6.

PR #33, opened 2019-03-04
Add completion file for myrepos (mr)

orel1:
Hello,
Here you can find a new zsh completion script for the command "mr" provided by [myrepos](http://myrepos.branchable.com/).


okapia:
Merged in #3908987. Thanks for the contribution.

I tweaked the function very slightly: the arguments array should have been declared local and it is wise to fallback on _default if a sub command is not found.

PR #34, opened 2019-03-10
KDEConnect autocompletion update

mtijink:
Use a device list with description (name and paired status) as autocompletion on newer KDE Connect versions. Also add the new option `--photo` and improve `--my-id`'s description.

I couldn't find zsh's policy on including unreleased options, so if I need to wait until a KDE Connect release until submitting this, please let me know.

Regarding the device autocomplete list with description, there might be some better way to do this. I developed the feature to output these autocompletions, so I can change it to something better (see https://invent.kde.org/kde/kdeconnect-kde/merge_requests/28). `kdeconnect-cli --zsh-device-autocomplete` outputs something like:
```
dd7972ee24a98fb5[G4 (paired)]
(more lines like that)
```

### DIFF
--- a/Completion/X/Command/_kdeconnect
+++ b/Completion/X/Command/_kdeconnect
@@ -1,7 +1,9 @@
 #compdef kdeconnect-cli
 
 _kdeconnect_device-ids() {
-  compadd $(kdeconnect-cli --list-available --id-only)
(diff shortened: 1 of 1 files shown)


mtijink:
At KDE Connect, we were also wondering if it would be better to ship this autocomplete file ourselves? Or would that cause all kinds of problems?

PR #35, opened 2019-05-05
Fix _gem push completion

spk:
Hi, I could reproduce the issue described in https://github.com/rubygems/rubygems/issues/2758 with zsh `5.7.1` and gem `2.5.2.1` using same arg as in `build` fixed the problem, Cheers


okapia:
Merged in #191e05b with a slight tweak to the change. Thanks for the contribution.

PR #36, opened 2019-06-24
Add completion file for GNU Stow

orel1:
Hello,
Here you can find a new zsh completion script for [GNU Stow](https://www.gnu.org/software/stow/).

### DIFF
--- a/Completion/Unix/Command/_stow
+++ b/Completion/Unix/Command/_stow
@@ -5,11 +5,8 @@
 #
 
 __stow_complete_packages() {
-  local stow_dir
-  local -a stow_pkg_list
-
-  eval set -A stow_dir $1
-  [[ -n $stow_dir ]] && stow_pkg_list=( $stow_dir/*(-/N:t) )
+  local stow_dir=${(Q)1}
+  local -a stow_pkg_list=( $stow_dir/*(-/N:t) )
 
   if [[ ${#stow_pkg_list} -gt 0 ]]; then
     _values -C "packages from $stow_dir" ${stow_pkg_list[@]}
--- a/Completion/Unix/Command/_stow
+++ b/Completion/Unix/Command/_stow
@@ -26,8 +26,8 @@ case $service in
     arguments=(
       '(- *)'{--help,-h}'[show help]'
       '(- *)'{--version,-V}'[show version number]'
-      '(-d --dir)'{-d+,--dir=}'[set the stow dir (default is current dir)]:stow dir:_files -/'
-      '(-t --target)'{-t+,--target=}'[set the target dir (default is parent of stow dir)]:target dir:_files -/'
+      '(-d --dir)'{-d+,--dir=}'[set the stow dir (default is current dir)]:stow dir [$PWD]:_files -/'
+      '(-t --target)'{-t+,--target=}'[set the target dir (default is parent of stow dir)]:target dir [../$PWD]:_files -/'
       # Several distinct actions can be specified in a single invocation
       # of the stow command (stow/unstow/restow). However, neither the
       # stow command nor this script will prevent you from using
--- a/Completion/Unix/Command/_stow
+++ b/Completion/Unix/Command/_stow
@@ -5,9 +5,12 @@
 #
 
 __stow_complete_packages() {
-  local IFS=$'\n'
-  local basedir="${1/\~/$HOME}"
-  _values -C "packages from $basedir" ${$( find -L "$basedir" -mindepth 1 -maxdepth 1 \( -name .git -o -name .svn \) -prune -o -type d -print 2>/dev/null | sed "s#^${basedir}/##" | sort -u ):-""}
+  local stow_pkg_list=( $1/*(-/N:t) )
+  if [[ ${#stow_pkg_list} -gt 0 ]]; then
+    _values -C "packages from $1" ${stow_pkg_list[@]}
+  else
+    _message "no packages found in $1"
+  fi
 }
 
 _stow(){
--- a/Completion/Unix/Command/_stow
+++ b/Completion/Unix/Command/_stow
@@ -21,7 +21,7 @@ _stow(){
 
   arguments=(
     '(- *)'{--help,-h}'[show help]'
-    '(- *)'{--version,-v}'[show version number]'
+    '(- *)'{--version,-V}'[show version number]'
     '(-d --dir)'{-d+,--dir=}'[set the stow dir (default is current dir)]:stow dir:_files -/'
     '(-t --target)'{-t+,--target=}'[set the target dir (default is parent of stow dir)]:target dir:_files -/'
     '*'{-S,--stow}'[stow the package names that follow]:stow package:->stow_package'
--- a/Completion/Unix/Command/_stow
+++ b/Completion/Unix/Command/_stow
@@ -13,79 +13,75 @@ __stow_complete_packages() {
   fi
 }
 
-_stow(){
-  local context state line curcontext="$curcontext" ret=1
-  typeset -A opt_args
-  # Others local variables
-  local stow_dir arguments
+case $service in
+  stow)
+    local context state line curcontext="$curcontext" ret=1
+    typeset -A opt_args
+    # Others local variables
+    local stow_dir arguments
 
-  arguments=(
-    '(- *)'{--help,-h}'[show help]'
-    '(- *)'{--version,-V}'[show version number]'
-    '(-d --dir)'{-d+,--dir=}'[set the stow dir (default is current dir)]:stow dir:_files -/'
-    '(-t --target)'{-t+,--target=}'[set the target dir (default is parent of stow dir)]:target dir:_files -/'
-    '*'{-S,--stow}'[stow the package names that follow]:stow package:->stow_package'
-    '*'{-D,--delete}'[unstow the package names that follow]:stow package:->stow_package'
-    '*'{-R,--restow}'[restow (unstow and stow again) the package names that follow]:stow package:->stow_package'
-    '--adopt[adopt already existing plain file]'
-    '--ignore=[ignore files ending with this perl regex]:regexp:'
-    "--defer=[don't stow files beginning with this perl regex]:regexp:"
-    '--override=[force stowing files beginning with this perl regex]:regexp:'
-    '--no-folding[disable any further tree folding or tree refolding]'
-    '(-p --compat)'{-p,--compat}'[use legacy algorithm for unstowing]'
-    '(-n -no --simulate)'{-n,--no,--simulate}'[do not actually make any filesystem changes]'
-    '(-v --verbose)'{-v,--verbose}'[increase or set the level of verbosity (from 0 to 5)]'
-    '*:stow package:->stow_package'
-  )
+    arguments=(
+      '(- *)'{--help,-h}'[show help]'
+      '(- *)'{--version,-V}'[show version number]'
+      '(-d --dir)'{-d+,--dir=}'[set the stow dir (default is current dir)]:stow dir:_files -/'
+      '(-t --target)'{-t+,--target=}'[set the target dir (default is parent of stow dir)]:target dir:_files -/'
+      '*'{-S,--stow}'[stow the package names that follow]:stow package:->stow_package'
+      '*'{-D,--delete}'[unstow the package names that follow]:stow package:->stow_package'
+      '*'{-R,--restow}'[restow (unstow and stow again) the package names that follow]:stow package:->stow_package'
+      '--adopt[adopt already existing plain file]'
+      '--ignore=[ignore files ending with this perl regex]:regexp:'
+      "--defer=[don't stow files beginning with this perl regex]:regexp:"
+      '--override=[force stowing files beginning with this perl regex]:regexp:'
+      '--no-folding[disable any further tree folding or tree refolding]'
+      '(-p --compat)'{-p,--compat}'[use legacy algorithm for unstowing]'
+      '(-n -no --simulate)'{-n,--no,--simulate}'[do not actually make any filesystem changes]'
+      '(-v --verbose)'{-v,--verbose}'[increase or set the level of verbosity (from 0 to 5)]'
+      '*:stow package:->stow_package'
+    )
 
-  if (( ${+commands[stow]} )) ; then
-    if [[ "$( _call_program version stow --version 2>/dev/null )" =~ "stow\ \(GNU\ Stow\)\ version\ ([0-9]+\.[0-9]+\.[0-9]+)" ]]; then
-      autoload -Uz is-at-least
-      if is-at-least 2.3.0 "$match[1]" ; then
-        arguments+=( '--dotfiles[enable special handling for dotfiles]' )
+    if (( ${+commands[stow]} )) ; then
+      if [[ "$( _call_program version stow --version 2>/dev/null )" =~ "stow\ \(GNU\ Stow\)\ version\ ([0-9]+\.[0-9]+\.[0-9]+)" ]]; then
+        autoload -Uz is-at-least
+        if is-at-least 2.3.0 "$match[1]" ; then
+          arguments+=( '--dotfiles[enable special handling for dotfiles]' )
+        fi
       fi
     fi
-  fi
 
-  _arguments -s -W -C $arguments && ret=0
+    _arguments -s -W -C $arguments && ret=0
 
-  case $state in
-    (stow_package)
-      if (( $+opt_args[-d] )) ; then
-        stow_dir="$opt_args[-d]"
-      elif (( $+opt_args[--dir] )) ; then
-        stow_dir="$opt_args[--dir]"
-      elif [[ ${(t)STOW_DIR} == *export* ]]; then
-        # if not provided from the command line, for the stow command, the stow
-        # directory is assumed to be the value of the "STOW_DIR" environment
-        # variable...
-        stow_dir="$STOW_DIR"
-      else
-        # ...if unset, the stow directory is assumed to be the current directory
-        stow_dir="$PWD"
-      fi
+    case $state in
+      (stow_package)
+        if (( $+opt_args[-d] )) ; then
+          stow_dir="$opt_args[-d]"
+        elif (( $+opt_args[--dir] )) ; then
+          stow_dir="$opt_args[--dir]"
+        elif [[ ${(t)STOW_DIR} == *export* ]]; then
+          # if not provided from the command line, for the stow command, the stow
+          # directory is assumed to be the value of the "STOW_DIR" environment
+          # variable...
+          stow_dir="$STOW_DIR"
+        else
+          # ...if unset, the stow directory is assumed to be the current directory
+          stow_dir="$PWD"
+        fi
 
-      __stow_complete_packages "$stow_dir"
+        __stow_complete_packages "$stow_dir"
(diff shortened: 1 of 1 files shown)


Completion/Unix/Command/_stow
```diff
@@ -0,0 +1,83 @@
+#compdef stow chkstow
+
+#
+# A zsh completion script for GNU stow (https://www.gnu.org/software/stow/)
+#
+
+__stow_complete_packages() {
+  local IFS=$'\n'
+  local basedir="${1/\~/$HOME}"
+  _values -C "packages from $basedir" ${$( find -L "$basedir" -mindepth 1 -maxdepth 1 \( -name .git -o -name .svn \) -prune -o -type d -print 2>/dev/null | sed "s#^${basedir}/##" | sort -u ):-""}
```
danielshahaf: Please avoid `find`, for performance on cygwin and for correctness (filenames with spaces, etc).  In this case:

```zsh
_values -c "..." $basedir/*(-/N:t)
```

Completion/Unix/Command/_stow
```diff
@@ -0,0 +1,83 @@
+#compdef stow chkstow
+
+#
+# A zsh completion script for GNU stow (https://www.gnu.org/software/stow/)
+#
+
+__stow_complete_packages() {
+  local IFS=$'\n'
+  local basedir="${1/\~/$HOME}"
```
danielshahaf: This seems like it's going to DTWT on `~foo` and `foo~bar`.  Could you please make it more robust?  For that matter, why isn't `local basedir="$1"` good enough?

Completion/Unix/Command/_stow
```diff
@@ -0,0 +1,83 @@
...
+_stow(){
+  local context state line curcontext="$curcontext" ret=1
+  typeset -A opt_args
+  # Others local variables
+  local stow_dir arguments
+
+  arguments=(
+    '(- *)'{--help,-h}'[show help]'
+    '(- *)'{--version,-v}'[show version number]'
+    '(-d --dir)'{-d,--dir=-}'[set the stow dir (default is current dir)]:stow dir:_files -/'
```
danielshahaf: Is `{-d,--dir=-}` correct?  GNU programs generally use `{-d+,--dir=}` (added plus, removed minus).

Completion/Unix/Command/_stow
```diff
@@ -0,0 +1,83 @@
...
+    '(-p --compat)'{-p,--compat}'[use legacy algorithm for unstowing]'
+    '(-n -no --simulate)'{-n,--no,--simulate}'[do not actually make any filesystem changes]'
+    '(-v --verbose)'{-v,--verbose}'[increase verbosity]'
+    '(-v --verbose)'--verbose=-'[set verbose level to n]:level:(o 1 2 3 4 5)'
+    '*:stow package:->stow_package'
+  )
+
+  if (( ${+commands[stow]} )) ; then
+    if [[ "$( _call_program version stow --version 2>/dev/null )" =~ "stow\ \(GNU\ Stow\)\ version\ ([0-9]+\.[0-9]+\.[0-9]+)" ]]; then
+      autoload -z is-at-least
```
danielshahaf: `autoload -Uz` ?

Completion/Unix/Command/_stow
orel1: Great, thanks for that! I didn't like this `find` too, but I couldn't find the _zsh way_ to do that.

Completion/Unix/Command/_stow
orel1: I don't know exactly why I had this, but with `_values -C "packages from $basedir" $basedir/*(-/N:t)`, `local basedir="$1"` seems sufficient. So, I'm going to make a modification in that way.

Completion/Unix/Command/_stow
orel1: You're right, I'll fix that and do the same for `-t,--target=-`

Completion/Unix/Command/_stow
orel1: indeed, according to man zshmisc _This is recommended for the use of functions supplied with the zsh distribution_ :disappointed_relieved:

Completion/Unix/Command/_stow
danielshahaf: You're welcome. It's called [
"glob qualifiers" ([documented in zshexpn](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Glob-Qualifiers)).

danielshahaf: Just noticed that I wrote `-c` in my previous comment. For avoidance of doubt, that was a typo: I meant to write `-C`, with a capital C as in your original.

orel1:
Following the previous discussions, I made some modifications/corrections.

it works, except on one point, if I use `--dir=$HOME/somedir` or `--dir=~/somedir` no completion results are found. For `~` this is really not a problem, but it seems that `$HOME` is not expand...

Completion/Unix/Command/_stow
danielshahaf: There are still a few instances of `--foo=-` in _stow and _chkstow.  Could you please confirm that they're correct, or change them if they aren't?

Completion/Unix/Command/_stow
```diff
@@ -55,7 +55,13 @@ _stow(){
         stow_dir="$opt_args[-d]"
       elif (( $+opt_args[--dir] )) ; then
         stow_dir="$opt_args[--dir]"
+      elif (( $+STOW_DIR )) ; then
+        # if not provided from the command line, for the stow command, the stow
+        # directory is assumed to be the value of the "STOW_DIR" environment
+        # variable...
+        stow_dir="$STOW_DIR"
```
danielshahaf: I wonder if it would be worthwhile to check `[[ ${(t)STOW_DIR} == *export* ]]`, in order to prevent mysterious behaviour for users who set the variable but forget to export it.

danielshahaf:
> it works, except on one point, if I use `--dir=$HOME/somedir` or `--dir=~/somedir` no completion results are found. For `~` this is really not a problem, but it seems that `$HOME` is not expand...

A tilde would be expanded there if MAGIC_EQUAL_SUBST were set.

I'm afraid I don't remember how this is usually handled.  You might try asking on the mailing list — it has more answerers than our github/gitlab presence.

Completion/Unix/Command/_stow
orel1: Argh, good point.

orel1:
ok. Thanks you for the feedback.
I'll take my chances on that side.

Completion/Unix/Command/_stow
```diff
@@ -0,0 +1,89 @@
+#compdef stow chkstow
+
+#
+# A zsh completion script for GNU stow (https://www.gnu.org/software/stow/)
+#
+
+__stow_complete_packages() {
```
okdana: Consider guarding this with `(( $+functions[__stow_complete_packages] )) ||`, so that the function doesn't get re-defined if it already exists. Also, the `complete` in the function name feels a bit redundant

Completion/Unix/Command/_stow
```diff
@@ -0,0 +1,89 @@
...
+  local stow_dir=${(Q)1}
+  local -a stow_pkg_list=( $stow_dir/*(-/N:t) )
+
+  if [[ ${#stow_pkg_list} -gt 0 ]]; then
+    _values -C "packages from $stow_dir" ${stow_pkg_list[@]}
```
okdana: You need to escape special characters in the operands if you're going to use `_values`; something like: `${${stow_pkg_list//\\/\\\\}//:/\\:}`

Alternatively you could replace all of this by `_wanted` and `compadd`, but you might have to change the way you call the function from `_arguments`

Also, descriptions should be singular (*package*, not *packages*)

Completion/Unix/Command/_stow
```diff
@@ -0,0 +1,89 @@
...
+  stow)
+    local context state line curcontext="$curcontext" ret=1
+    typeset -A opt_args
+    # Others local variables
+    local stow_dir arguments
+
+    arguments=(
+      '(- *)'{--help,-h}'[show help]'
+      '(- *)'{--version,-V}'[show version number]'
+      '(-d --dir)'{-d+,--dir=}'[set the stow dir (default is current dir)]:stow dir:_files -/'
```
okdana: This doesn't seem to be in the completion style guide, but either Oliver or Daniel once suggested to me to put defaults in square brackets at the end of the match description, as in `stow directory [$PWD]`

Completion/Unix/Command/_stow
```diff
@@ -0,0 +1,89 @@
...
+      '(-t --target)'{-t+,--target=}'[set the target dir (default is parent of stow dir)]:target dir:_files -/'
+      '*'{-S,--stow}'[stow the package names that follow]:stow package:->stow_package'
```
okdana: Looking at the `stow` man page, i don't think these specs are right:

https://www.gnu.org/software/stow/manual/stow.html#Mixing-Operations

Off the top of my head i'm not sure what would be the ideal way to do this. You could use the `:*:...` mechanism, but there's no way to return from it. idk. At the least, you might consider leaving a note here that it's not accurate

Also, the `:stow package:` in all of these is superfluous (and misleading)

Completion/Unix/Command/_stow
```diff
@@ -0,0 +1,89 @@
...
+      '*'{-D,--delete}'[unstow the package names that follow]:stow package:->stow_package'
+      '*'{-R,--restow}'[restow (unstow and stow again) the package names that follow]:stow package:->stow_package'
+      '--adopt[adopt already existing plain file]'
+      '--ignore=[ignore files ending with this perl regex]:regexp:'
+      "--defer=[don't stow files beginning with this perl regex]:regexp:"
+      '--override=[force stowing files beginning with this perl regex]:regexp:'
+      '--no-folding[disable any further tree folding or tree refolding]'
+      '(-p --compat)'{-p,--compat}'[use legacy algorithm for unstowing]'
+      '(-n -no --simulate)'{-n,--no,--simulate}'[do not actually make any filesystem changes]'
+      '(-v --verbose)'{-v,--verbose}'[increase or set the level of verbosity (from 0 to 5)]'
```
okdana: It looks like `--verbose` is supposed to take an optional verbosity level, and it should be possible to use them multiple times so they stack up (e.g., you can do either `-vvv` or `--verbose=3`)

Completion/Unix/Command/_stow
```diff
@@ -0,0 +1,89 @@
...
+      '*:stow package:->stow_package'
+    )
+
+    if (( ${+commands[stow]} )) ; then
```
okdana: Usually we don't worry about version checks like this, we just support whatever the latest options are. Especially since it's just this one option, i wouldn't bother with it personally. (If you *did* keep it, i'd try to use `_pick_variant` or something rather than making an external call every single time)

Completion/Unix/Command/_stow
```diff
@@ -0,0 +1,89 @@
...
+
+    _arguments -s -W -C $arguments && ret=0
+
+    case $state in
+      (stow_package)
+        if (( $+opt_args[-d] )) ; then
+          stow_dir="$opt_args[-d]"
+        elif (( $+opt_args[--dir] )) ; then
+          stow_dir="$opt_args[--dir]"
+        elif [[ ${(t)STOW_DIR} == *export* ]]; then
```
okdana: Should this maybe check for emptiness?

Completion/Unix/Command/_stow
```diff
@@ -0,0 +1,89 @@
...
+          # if not provided from the command line, for the stow command, the stow
+          # directory is assumed to be the value of the "STOW_DIR" environment
+          # variable...
+          stow_dir="$STOW_DIR"
+        else
+          # ...if unset, the stow directory is assumed to be the current directory
+          stow_dir="$PWD"
+        fi
+
+        __stow_complete_packages "$stow_dir"
```
okdana: You're not setting `ret` here

Completion/Unix/Command/_stow
```diff
@@ -0,0 +1,89 @@
...
+    if (( ${+commands[stow]} )) ; then
+      if [[ "$( _call_program version stow --version 2>/dev/null )" =~ "stow\ \(GNU\ Stow\)\ version\ ([0-9]+\.[0-9]+\.[0-9]+)" ]]; then
+        autoload -Uz is-at-least
+        if is-at-least 2.3.0 "$match[1]" ; then
+          arguments+=( '--dotfiles[enable special handling for dotfiles]' )
+        fi
+      fi
+    fi
+
+    _arguments -s -W -C $arguments && ret=0
```
okapia: Are you sure about the use of `-W` here. Genuine cases where `-w` or `-W` are needed are very rare so I tend not to believe it when I see it used.

Completion/Unix/Command/_stow
```diff
@@ -0,0 +1,89 @@
...
+  if [[ ${#stow_pkg_list} -gt 0 ]]; then
+    _values -C "packages from $stow_dir" ${stow_pkg_list[@]}
+  else
+    _message "no packages found in $stow_dir"
+  fi
+}
+
+case $service in
+  stow)
+    local context state line curcontext="$curcontext" ret=1
```
okapia: Making context local is superfluous. It is nearly always fine to stick with curcontext abd pass -C to _arguments/_values wherever you use states. context is an array and is only applicable where multiple states could be active together (which only happens with things like optional arguments).

okapia:
It'd be great to get this merged into zsh.

Thanks for addressing the initial points from Daniel's review. And sorry that it was then left for a couple of months before further review comments were added. It'd still be good to address these further points. Can we perhaps get this finalised? Thanks

orel1:
I agree with that, but time has gone by far too quickly in recent weeks....

I'm trying to respond to the last comments this weekend.

Completion/Unix/Command/_stow
orel1: I don't see how to do that... could you give me an example please ?

Completion/Unix/Command/_stow
orel1: Indeed, I hadn't thought of this case. But after a few tests, this doesn't seem to be a problem for stow, even if it can lead to surprising results... I have added a note to explain that point.

However, I do not understand the following comment:

> Also, the :stow package: in all of these is superfluous (and misleading)

if I remove `:stow package:` it doesn't work anymore...

Completion/Unix/Command/_stow
danielshahaf: Change `:stow dir:` to `:stow dir [$PWD]:`.

One of us will add this guidance to Etc/completion-style-guide (or you can do that, of course).

Completion/Unix/Command/_stow
orel1: Yes, initially I used this (see my first commit 64fcfa8f6) :

    '(-v --verbose)'{-v,--verbose}'[increase verbosity]'
    '(-v --verbose)'--verbose=-'[set verbose level to n]:level:(0 1 2 3 4 5)'`

I don't know if there's a better way to do this

Completion/Unix/Command/_stow
orel1: Just to be sure, are you thinking about the case where the exported variable is empty ?

Completion/Unix/Command/_stow
orel1: Honestly, the use of the ret variable is not very clear to me....
I should add  `&& ret=0` at the end of the line ?

Completion/Unix/Command/_stow
orel1: done in 3570f596a

Completion/Unix/Command/_stow
orel1: done in 3c75ffcc4

Completion/Unix/Command/_stow
orel1: done in 8ae64abcd

Completion/Unix/Command/_stow
orel1: done in b9c086fb5

Completion/Unix/Command/_stow
orel1: done in 6178a88d4

Completion/Unix/Command/_stow
okapia: Try just `: :`.

The description is ignored because it goes to the stow_package state.

Completion/Unix/Command/_stow
okapia: If the argument to --verbose is optional, use two colons to indicate so. For example:

    '*-v[increase verbosity]'
    '*--verbose=-[increase verbosity]::level:(0 1 2 3 4 5)'

I'd be inclined to have the same description so the options are grouped together. Things like "to n" seem weird without more context for what "n" is anyway. And if options can be repeated, we want the initial `*` rather than exclusion lists.

Completion/Unix/Command/_stow
okapia: I can't speak for okdana but I would assume so.

Completion/Unix/Command/_stow
okapia: Completion functions should return 0 if they added matches or did something useful. and 1 if not. Returning 1 can mean that zsh goes on to try further "completers" - approximate completion and so on.

A common way to do this is to start the script by declaring `ret=1`, put `&& ret=0` after ever call to compadd or other completion functions and then put `return ret` at the end. If you use this approach, you need to remember the `&& ret=0` on every call that might add some completion candidates.

Completion/Unix/Command/_stow
okdana: Yeah, that's what i meant. I'm not sure how Stow actually handles that — i know some environment variables are only checked for set-ness — but intuitively i'd assume that `STOW_DIR=''` is treated as unset

Completion/Unix/Command/_stow
danielshahaf: Done by @okdana (thanks!) in [workers/44969](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=44969).

Completion/Unix/Command/_stow
danielshahaf: It might be treated as equivalent to cwd.  For example, `s=""; echo $s:P` works this way.

Completion/Unix/Command/_stow
orel1: done in 0ebc9639e

Completion/Unix/Command/_stow
orel1: done in 675da0869

Completion/Unix/Command/_stow
orel1: done in b1272e3e8

Completion/Unix/Command/_stow
orel1: done in 414270e16

Completion/Unix/Command/_stow
orel1: done in a3e659bdc

orel1:
Thank you for all your answers :smile:

okapia:
Merged in 058cc10.

Thanks for the contribution.

I'm not sure if any of the normal zsh developers use GNU stow at all. I certainly don't. So please help keep the completion current by sending us fixes as and when there are updates to stow. Thanks

PR #37, opened 2019-08-31
Make it easier to understand what "fc -e" does.

nedbat:


Doc/Zsh/builtins.yo
```diff
@@ -156,7 +156,8 @@ first referenced.
 
 If var(name) consists of an absolute path, the function is defined to
 load from the file given (searching as usual for dump files in the given
-location).  The name of the function is the basename (non-directory
+location).  The name of the function is the bas
```
roosemberth: This looks like an accidental edit.

nedbat: Yes, sorry.

nedbat: fixed

Doc/Zsh/builtins.yo
```diff
@@ -768,7 +768,7 @@ However, if the current event has added entries to the history with
 includes all new history entries since the current event began.
 
 When the tt(-l) flag is given, the resulting events are listed on
-standard output.  Otherwise the editor program var(ename) is invoked on a
+standard output.  Otherwise the editor program specified by tt(-e) var(ename) is invoked on a
```
danielshahaf: I suggest to add commas around `var(ename)`.

I assume the point of the change is to add the string «-e» somewhere in the help so it can more easily be grepped for.  +1 to that.

Doc/Zsh/builtins.yo
danielshahaf: @roosemberth Thanks for the review!

Doc/Zsh/builtins.yo
nedbat: That's exactly the goal :)

nedbat: I like it better without the commas, so that "-e ename" appears in the text the same way it does in the summary at the top of the section.

danielshahaf: I think the grammar are needed for syntax reasons (English syntax, not yodl). The greppability goal is achieved by adding the option name.

danielshahaf: I'm used to discussing "the `-e` option", not *"the `-e` _ename_ option". Furthermore, I don't generally expect synopsis and prose to use precisely the same words and markup.

All that said, I'll merge whatever you write so long as it's correct syntax. I don't overrule patch submitters :-)

nedbat: I don't think it is incorrect syntax as it stands.  It certainly reads naturally to me.

danielshahaf: Fair enough. Merged in 63cc5279b38bb07acd8074dcfcd0b0c354e3dfc9. Cheers!

danielshahaf:
Merged.

PR #38, opened 2019-09-18
fix completion text for git update-index --chmod

tschulte:


danielshahaf:
Thanks for the patch! Merged in bed4e9168.

PR #39, opened 2019-09-22
Update notmuch zsh completions

cbcoutinho:


Completion/Unix/Command/_notmuch
```diff
@@ -1,80 +1,292 @@
-#compdef notmuch
+#compdef notmuch -p notmuch-*
 
-_notmuch_commands()
-{
+# ZSH completion for `notmuch`
+# Copyright © 2018 Vincent Breitmoser <look@my.amazin.horse>
```
danielshahaf: Under what license does Mr Breitmoser place the code?

Completion/Unix/Command/_notmuch
```diff
@@ -1,80 +1,292 @@
...
+    '*--include=[configure metadata to output (default all)]:metadata type:(config properties tags)' \
+    '--output=[write output to file]:output file:_files' \
+    '*::search term:_notmuch_search_term'
+}
+
+_notmuch_new() {
+  _arguments \
+    '--no-hooks[prevent hooks from being run]' \
+    '--quiet[do not print progress or results]' \
+    '--full-scan[don''t rely on directory modification times for scan]' \
```
danielshahaf: The `don't` is incorrectly escaped: cshjunkiequotes syntax can't be used here.

Completion/Unix/Command/_email-notmuch
```diff
@@ -0,0 +1,9 @@
+#autoload
```
danielshahaf: Should be in a different directory: Types/ rather than Command/.

Completion/Unix/Command/_email-notmuch
```diff
@@ -0,0 +1,9 @@
...
+
+local expl
+local -a notmuch_addr
+
+notmuch_addr=( ${(f)"$(notmuch address --deduplicate=address --output=address -- from:/$PREFIX/)"} )
```
danielshahaf: Does $PREFIX need to be shell- or regexp-escaped?

Please look into `_call_program` and use it where needed, to allow overriding the commands using zstyle's.

Completion/Unix/Command/_notmuch
cbcoutinho: Notmuch is GPL-3, based on the COPYING file in the project root [0]. I'm only assuming this applies to the zsh completions as well

https://github.com/notmuch/notmuch/blob/master/COPYING

Completion/Unix/Command/_email-notmuch
cbcoutinho: I can move it, but where in the zsh docs does it explain what these different directories mean?

I was only able to find this [0]

http://zsh.sourceforge.net/Doc/Release/Completion-System.html#Completion-Directories

Completion/Unix/Command/_notmuch
cbcoutinho: so either `don\'t` or `do not` will be ok?

Completion/Unix/Command/_notmuch
okapia: We don't just pull in GPL-3 licenced code into zsh, even if it is in a self-contained file like a completion function. While we can do this in theory, it forces a big pile or bureaucratic work on packagers for some distributions if zsh comes under a mix of licences. If notmuch includes a zsh completion, we should simply remove the _notmuch from the zsh distribution.

Completion/Unix/Command/_notmuch
okapia: In general, I would recommend "don't" (with double quotes around the whole line instead of single).

danielshahaf: Contractions are fine, but `'foo''bar'` is equivalent to `'foobar'`. In this case I'd recommend `"--full-scan[don't]"`.

Completion/Unix/Command/_email-notmuch
danielshahaf: > where in the zsh docs does it explain what these different directories mean?

Nowhere :(

cbcoutinho: :( indeed

danielshahaf: I'd add it to the link you posted, but I don't think the user manual is the right place to document the source code layout in.  I'm not sure where the information would fit better, either.  Where would you expect that bit of info to be documented?

cbcoutinho: Looking at the user documentation I was trying to answer the question:

If a package manager is used to build and install packages (thus not locally in `/usr/local`), where should they put the zsh completions? I'm assuming that /usr/local/zsh/functions is probably for internal completions for zsh itself, but packages seem to install their completions into those subdirectories anyways - at least that's the case on my system now that I've been looking at it.

Do packages install their zsh completions into the /usr/share/zsh/{functions,site-functions,vendor-completions,???}.

That isn't stated in the docs, but might be useful for package managers

Completion/Unix/Command/_notmuch
cbcoutinho: I could also delete this PR and send a new one just removing these files from the zsh distribution

Completion/Unix/Command/_email-notmuch
danielshahaf: > packages seem to install their completions into those subdirectories anyways - at least that's the case on my system now that I've been looking at it.

This is controlled by the `--enable-function-subdirs` configure flag.

> Do packages install their zsh completions into the /usr/share/zsh/{functions,site-functions,vendor-completions,???}.

`site-functions` is reserved to the local sysadmin. `vendor-completions` is probably added to fpath by your distro as a place for distro packages to put their scripts in (that's how Debian does it, anyway). `$PREFIX/share/zsh/$ZSH_VERSION/functions` is where `make install` places the functions shipped with zsh itself.  Does your use-case fall into one of these?

This is a recurring question — the debian zsh packagers get it too from time to time — so we should probably add the answer to the docs.

Completion/Unix/Command/_notmuch
okapia: 'There is an rc_quotes option which allows 'foo''bar' to be foo'bar but you can't rely on user options in completion functions.

Completion/Unix/Command/_email-notmuch
okapia: `vendor-completions` is purely a Debian/Ubuntu thing.

Normally packages install their functions into `$prefix/share/zsh/site-functions`. That is at least going to be in the default $fpath for zsh users. It is probably up to distributions to separate things out.

okapia:
_notmuch has now been removed in ea78d0d
Thanks anyway.

PR #40, opened 2019-10-26
adjust zfs completions so that it uses acltype on linux instead of aclmode

tcheinen:
zfs doesn't have an aclmode option on linux.  I set the zfs completions to use acltype instead on a linux system.

### DIFF
--- a/Completion/Unix/Command/_zfs
+++ b/Completion/Unix/Command/_zfs
@@ -145,7 +145,6 @@ _zfs() {
 	# set the sorting for *size properties to false by default?
 	rw_properties=(
 		"aclinherit:value:(discard noallow restricted passthrough passthrough-x)"
-		$(if [[ "$OSTYPE" == "linux-gnu" ]]; then echo "acltype:value:(off noacl posixacl)"; else echo "aclmode:value:(discard mask passthrough)"; fi );
 		"atime:value:(on off)"
 		"canmount:value:(on off noauto)"
 		"checksum:value:(on off fletcher2 fletcher4 sha256 sha256+mac)"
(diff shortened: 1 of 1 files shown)


Completion/Unix/Command/_zfs
```diff
@@ -185,6 +185,7 @@ _zfs() {
 		$share_rw_properties
 	)
 
+
```
okapia: not sure if this was intentional but I don't think the extra blank line is needed.

Completion/Unix/Command/_zfs
```diff
@@ -145,7 +145,7 @@ _zfs() {
 	# set the sorting for *size properties to false by default?
 	rw_properties=(
 		"aclinherit:value:(discard noallow restricted passthrough passthrough-x)"
-		"aclmode:value:(discard mask passthrough)"
+		$(if [[ "$OSTYPE" == "linux-gnu" ]]; then echo "acltype:value:(off noacl posixacl)"; else echo "aclmode:value:(discard mask passthrough)"; fi );
```
okapia: Thanks for the contribution. This is a useful improvement.

Using echo inside `$( ... )` creates a subshell and pipe. How about appending to rw_properties after the end of the assignment - something like:

    if [[ $OSTYPE = linux-gnu ]]; then
      rw_properties+=( "acltype:value:(off noacl posixacl)" )
    else
      rw_properties+=( ... )
    fi

Actually, while changing this, from a quick check it looks like the possible values are discard|groupmask|passthrough|restricted for openzfs (FreeBSD and Illumos) while the original form of the completion is right for Solaris. So it could be good to handle that difference.

Completion/Unix/Command/_zfs
tcheinen: it wasn't - actually I originally appended it but I swapped it to use an inline expression because I wasn't sure if order was important.  I'll fix this line when I swap back to appending it

okapia:
Merged in 5cf1d6d.

Thanks for the contribution.
I tweaked it a tiny bit for the Solaris handling. The $implementation variable from the earlier call to _pick_variant is a better way to detect openzfs (illumos) vs. oracle solaris. I think $OSTYPE is the same on both.

PR #41, opened 2019-10-29
Extend _dbus with dbus-launch completion

psprint:
Extension of the dbus completion with support of dbus-launch command. Follow up to https://github.com/zsh-users/zsh-completions/pull/661.


okapia:
That looks good. Merged in 2c94b59. Thanks for the contribution.

psprint:
N/p, thanks for merging.

PR #43, opened 2019-11-22
Fix markdown error and typo

LuckyWindsck:
I found some markdown errors in `./Etc/completion-style-guide` that quote back-ticks with single quotation marks, so I replace them all.

Also I typo fix in the same file.

### DIFF
--- a/Etc/completion-style-guide
+++ b/Etc/completion-style-guide
@@ -521,7 +521,7 @@ Misc. remarks
       since multiple arguments to the same option are represented in a
       colon-delimited fashion, backslashes and colons passed by the user
       are escaped. For instance, the option-arguments parsed from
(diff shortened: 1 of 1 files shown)


okapia:
Thanks for the contributions.

However, the completion-style-guide is not intended to be in markdown syntax. It would have been given a `.md`extension. The backquote and apostrophe are being used as opening/closing quotes which as a quoting style has the advantage of not being something you might confuse with real shell quoting. It might also be the influence of yodl which is the documentation system zsh uses that has led to this. Unicode quotes might be a reasonable choice nowadays, or perhaps converting the file properly to markdown. I'd want to pass that by the mailing list if it was to be changed.

I also think that the second patch, is not a typo but is intentional to show how opt_args is assigned both for the case of arguments that come after and in the same shell word as the option.

PR #44, opened 2020-01-07
Allow completion for picocom to list symlinks to character devices

Zenkibou:
The default picocom completion does not list symlinks to character devices.

Symlinks to character devices are useful to name devices, especially for usb-serial adapters where the name depends on the plug-in order.

This change allows the completion to list both the character devices and the symlinks to character devices.


danielshahaf:
Thanks, merged in 58ff4adbb0ee5ef09b50f7b62bda428006cd57d8 for 5.8.

PR #45, opened 2020-01-16
Add zstd file completion to tar

svenstaro:
zstd is a modern compression format that is gaining more support. For instance, Arch Linux recently switched its packages to .zstd.

### DIFF
--- a/Completion/Unix/Type/_tar_archive
+++ b/Completion/Unix/Type/_tar_archive
@@ -20,8 +20,10 @@ if [[ "$1" = *[urtx]* ]]; then
     _files "$expl[@]" -g '*.(tar|TAR).bz2(-.)'
   elif [[ "$1" = *J* ]]; then
     _files "$expl[@]" -g '*.(tar|TAR).(lzma|xz)(-.)'
+  elif [[ "$1" = *zstd* ]]; then
+    _files "$expl[@]" -g '*.(tar|TAR).(zst|zstd)(-.)'
   elif [[ "$_cmd_variant[$service]" == (gnu|libarchive) ]]; then
-    _files "$expl[@]" -g '*.((tar|TAR)(.gz|.GZ|.Z|.bz2|.lzma|.xz|)|(tbz|tgz|txz))(-.)'
+    _files "$expl[@]" -g '*.((tar|TAR)(.gz|.GZ|.Z|.bz2|.lzma|.xz|.zst|.zstd)|(tbz|tgz|txz))(-.)'
   else
     _files "$expl[@]" -g '*.(tar|TAR)(-.)'
   fi


Completion/Unix/Type/_tar_archive
```diff
@@ -20,8 +20,10 @@ if [[ "$1" = *[urtx]* ]]; then
     _files "$expl[@]" -g '*.(tar|TAR).bz2(-.)'
   elif [[ "$1" = *J* ]]; then
     _files "$expl[@]" -g '*.(tar|TAR).(lzma|xz)(-.)'
+  elif [[ "$1" = *zstd* ]]; then
+    _files "$expl[@]" -g '*.(tar|TAR).(zst|zstd)(-.)'
   elif [[ "$_cmd_variant[$service]" == (gnu|libarchive) ]]; then
-    _files "$expl[@]" -g '*.((tar|TAR)(.gz|.GZ|.Z|.bz2|.lzma|.xz|)|(tbz|tgz|txz))(-.)'
```
okapia: Have you verified with both libarchive and GNU tar? I only have the libarchive variant available to me and it appears to have a --zstd option with no short form.

Completion/Unix/Type/_tar_archive
```diff
@@ -20,8 +20,10 @@ if [[ "$1" = *[urtx]* ]]; then
...
+    _files "$expl[@]" -g '*.((tar|TAR)(.gz|.GZ|.Z|.bz2|.lzma|.xz|.zst|.zstd)|(tbz|tgz|txz|tzst|tzstd))(-.)'
```
okapia: Are `tzst` and and `tzstd` extensions really a thing or are you just making assumptions while copy/pasting? The original purpose behind `tgz`, `tbz` extensions was to cope with 8.3 filename limitations on old operating systems so they mostly are just a relic from the past. `tzst` doesn't even fit in 3 characters.

svenstaro: Oh, today I learned. I had no idea it was more than just an abbreviation for the heck of it. So to answer your question: It was just an assumption that this kind of abbreviation should be continued. However, I see now that it doesn't really make sense. I'll remove it.

Completion/Unix/Type/_tar_archive
svenstaro: Yeah, I checked: https://github.com/libarchive/libarchive/blob/2f3033ca23f8c21160506c3c7ac8a0df0d3fde42/cpio/bsdcpio.1#L189

danielshahaf: GNU tar supports --zstd too, with no short form either.

Completion/Unix/Type/_tar_archive
```diff
@@ -20,8 +20,10 @@ if [[ "$1" = *[urtx]* ]]; then
     _files "$expl[@]" -g '*.(tar|TAR).bz2(-.)'
   elif [[ "$1" = *J* ]]; then
     _files "$expl[@]" -g '*.(tar|TAR).(lzma|xz)(-.)'
+  elif [[ "$1" = *zstd* ]]; then
```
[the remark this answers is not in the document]
svenstaro: You're right. I thought I tested this locally but it was an old file. I had a look for an obvious way to implement what you said but I'm not really knowledgable enough about the structure of this whole completion structure to be honest.

okapia:
Looks like someone applied a similar patch separately covering .lzo too. Sorry, I should have got around to applying this quicker. I'll close this now. Thanks for the contribution. If you're unhappy with the state of the function in the current git repo, you can always open a fresh request.

svenstaro:
Eh, it's fine. If it works, it works.

PR #46, opened 2020-01-22
Better restric git restore file options

harogaston:
Limit the files suggested after --staged and --worktree flags

### DIFF
--- a/Completion/Unix/Command/_git
+++ b/Completion/Unix/Command/_git
@@ -1567,14 +1567,17 @@ _git-restore() {
 
   case $state in
     pathspecs)
-      if [[ -z ${opt_args[(I)-S|--staged]} ]]
+      integer opt_S opt_W
+      [[ -n ${opt_args[(I)-S|--staged]} ]] && opt_S=1
+      [[ -n ${opt_args[(I)-W|--worktree]} ]] && opt_W=1
+      if (( opt_S && opt_W ))
       then
-        __git_ignore_line __git_changed-in-working-tree_files
-      elif [[ -z ${opt_args[(I)-W|--worktree]} ]]
+        __git_ignore_line __git_changed_files && ret=0
+      elif (( opt_S ))
       then
-        __git_ignore_line __git_changed-in-index_files
+        __git_ignore_line __git_changed-in-index_files && ret=0
       else
-        __git_ignore_line __git_tree_files ${PREFIX:-.} ${(Qv)opt_args[(i)-s|--source]:-HEAD} && ret=0
+        __git_ignore_line __git_changed-in-working-tree_files && ret=0
       fi
     ;;
     sources)
--- a/Completion/Unix/Command/_git
+++ b/Completion/Unix/Command/_git
@@ -1567,18 +1567,14 @@ _git-restore() {
 
   case $state in
     pathspecs)
-      if [[ -z ${opt_args[(I)-s|--source|-S|--staged]} ]] &&
-	# use index as a default base unless -S is specified
-	__git_ignore_line __git_changed-in-working-tree_files
+      if [[ -z ${opt_args[(I)-S|--staged]} ]]
       then
-	ret=0
-      elif [[ -z ${opt_args[(I)-s|--source|-W|--worktree]} ]] &&
-	# use index as a default base unless -S is specified
-	__git_ignore_line __git_changed-in-index_files
+        __git_ignore_line __git_changed-in-working-tree_files
+      elif [[ -z ${opt_args[(I)-W|--worktree]} ]]
       then
-	ret=0
+        __git_ignore_line __git_changed-in-index_files
       else
-	__git_ignore_line __git_tree_files ${PREFIX:-.} ${(Qv)opt_args[(i)-s|--source]:-HEAD} && ret=0
+        __git_ignore_line __git_tree_files ${PREFIX:-.} ${(Qv)opt_args[(i)-s|--source]:-HEAD} && ret=0
       fi
     ;;
     sources)
--- a/Completion/Unix/Command/_git
+++ b/Completion/Unix/Command/_git
@@ -1569,7 +1569,12 @@ _git-restore() {
     pathspecs)
       if [[ -z ${opt_args[(I)-s|--source|-S|--staged]} ]] &&
 	# use index as a default base unless -S is specified
-	__git_ignore_line __git_modified_files
+	__git_ignore_line __git_changed-in-working-tree_files
+      then
+	ret=0
+      elif [[ -z ${opt_args[(I)-s|--source|-W|--worktree]} ]] &&
+	# use index as a default base unless -S is specified
+	__git_ignore_line __git_changed-in-index_files
       then
 	ret=0
       else
--- a/Completion/Unix/Command/_git
+++ b/Completion/Unix/Command/_git
@@ -1548,7 +1548,7 @@ _git-restore() {
   local -A opt_args
 
   _arguments -C -s -S $endopt \
-    '(-s --source)'{-s,--source}'[specify which tree-ish to checkout from]:source tree:->sources' \
+    '(-s --source)'{-s,--source=}'[specify which tree-ish to checkout from]:source tree:->sources' \
     '(-S --staged)'{-S,--staged}'[restore the index]' \
     '(-W --worktree)'{-W,--worktree}'[restore the working tree (default)]' \
     '--ignore-unmerged[ignore unmerged entries]' \


danielshahaf:
Sorry, too swamped to review this currently.  I'll add it to my list, but please someone beat me to it.

harogaston:
For a git repo with the following status:

```
~/T/fake-repo ‹master*› $ git ss                                                                                                                    
On branch master
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   staged-file

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   modified-file

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        new-file
```

the command will behave like:

```
$> git restore <TAB>
modified-file

$> git restore -S|--staged <TAB>
staged-file

$> git restore -W|--worktree <TAB>
modified-file

$> git restore -W|--worktree -S|--staged <TAB>
modified-file staged-file

$> git restore -W|--worktree -S|--staged modified-file <TAB>
staged-file

$> git restore -W|--worktree -S|--staged staged-file -s|--source <TAB>
HEAD    master
87ae776  -- [HEAD]    Initial commit (2 hours ago)
```

Completion/Unix/Command/_git
```diff
@@ -1567,13 +1567,14 @@ _git-restore() {
...
-      if [[ -z ${opt_args[(I)-s|--source|-S|--staged]} ]] &&
-	# use index as a default base unless -S is specified
-	__git_ignore_line __git_modified_files
+      if [[ -z ${opt_args[(I)-S|--staged]} ]]
       then
-	ret=0
+        __git_ignore_line __git_changed-in-working-tree_files
+      elif [[ -z ${opt_args[(I)-W|--worktree]} ]]
+      then
+        __git_ignore_line __git_changed-in-index_files
```
danielshahaf: Finally got back to this.  Sorry for the long delay.

I'm reviewing this against git-restore(1) from git 1:2.25.1-1 (Debian package) and have a few concerns:

1. The man page says the command is experimental and its behaviour may change.  This would seem to imply that the command's current behaviour is not guaranteed to be forward compatible with its future behaviour.  How should `_git-restore` handle this?  Perhaps it should default to doing nothing unless the user has opted in to completing experimental commands?  (The opt in will likely take the form of a zstyle setting.)

    This is not a regression, however, and as such not a blocker.

2. The `if` and `elif` branch should have `&& ret=0` appended to them.

3. It's possible to specify both -S and -W.  In that case, I assume that files modified _either_ in the index or in the worktree should be offered.  I think `__git_changed_files` will be the right function to use in this case.

Putting this all together, I think the following would be better:

    if --staged and --worktree:
        __git_ignore_line __git_changed_files && ret=0
    elif --staged:
        __git_ignore_line __git_changed-in-index_files && ret=0
    elif --worktree:
        __git_ignore_line __git_changed-in-working-tree_files && ret=0
    else:
        __git_ignore_line __git_tree_files ${PREFIX:-.} ${(Qv)opt_args[(i)-s|--source]:-HEAD} && ret=0

WDYT?

Apologies again for the delay.

danielshahaf:
Hello again @harogaston, could you test the following and confirm it offers everything it should offer?  It's your patch plus the above review points.  Thanks!

```diff
diff --git a/Completion/Unix/Command/_git b/Completion/Unix/Command/_git
index 9a44ddbbe..78ebdc5d7 100644
--- a/Completion/Unix/Command/_git
+++ b/Completion/Unix/Command/_git
@@ -1567,13 +1567,20 @@ _git-restore() {
 
   case $state in
     pathspecs)
-      if [[ -z ${opt_args[(I)-s|--source|-S|--staged]} ]] &&
-	# use index as a default base unless -S is specified
-	__git_ignore_line __git_modified_files
+      integer opt_S opt_W
+      [[ -z ${opt_args[(I)-S|--staged]} ]] && opt_S=1
+      [[ -z ${opt_args[(I)-W|--worktree]} ]] && opt_W=1
+      if (( opt_S && opt_W ))
       then
-	ret=0
+        __git_ignore_line __git_changed_files && ret=0
+      elif (( opt_S )) 
+      then
+        __git_ignore_line __git_changed-in-working-tree_files && ret=0
+      elif (( opt_W ))
+      then
+        __git_ignore_line __git_changed-in-index_files && ret=0
       else
-	__git_ignore_line __git_tree_files ${PREFIX:-.} ${(Qv)opt_args[(i)-s|--source]:-HEAD} && ret=0
+        __git_ignore_line __git_tree_files ${PREFIX:-.} ${(Qv)opt_args[(i)-s|--source]:-HEAD} && ret=0
       fi
     ;;
     sources)
```

harogaston:
Hi @danielshahaf sure, let me try that. I'm no expert on zsh scripting so take my contribution with a pinch of salt. I didn't know about the `ret=0` for example.
Sorry had a busy week.

danielshahaf:
> Hi @danielshahaf sure, let me try that. I'm no expert on zsh scripting so take my contribution with a pinch of salt. I didn't know about the `ret=0` for example.
> Sorry had a busy week.
> 

No worries, whenever you get around to it is fine.  The ret=0 isn't a generic zsh scripting thing; it's a completion system idiom, because completion functions are supposed to return non-zero if they didn't add matches.

harogaston:
Well I tried that, thank you for making that code a lot more readable. I took your ideas but made some modifications. Some things that didn't work for me which I fixed were:

- `git restore <tab>` was giving all modified files (staged and working tree) but should only be suggesting modified working tree files. You can even do a `git restore a-staged.file` and check that it does nothing. I take it that that is the correct behavior from the [documentation](https://git-scm.com/docs/git-restore#Documentation/git-restore.txt---staged):
> Specify the restore location. If neither option is specified, by default the working tree is restored. Specifying --staged will only restore the index. Specifying both restores both.

- `git restore -S -W` was giving all tree files but should only be suggesting modified files (staged and worktree) instead
- `git restore -S` and `git restore -T` were ok

```diff
diff --git a/home/gaston/Downloads/_git b/usr/share/zsh/functions/Completion/Unix/_git
index dc9c331357..ab4cfca32a 100644
--- a/usr/share/zsh/functions/Completion/Unix/_git
+++ b/usr/share/zsh/functions/Completion/Unix/_git
@@ -1567,14 +1567,17 @@ _git-restore() {

   case $state in
     pathspecs)
-      if [[ -z ${opt_args[(I)-S|--staged]} ]]
+      integer opt_S opt_W
+      [[ -n ${opt_args[(I)-S|--staged]} ]] && opt_S=1
+      [[ -n ${opt_args[(I)-W|--worktree]} ]] && opt_W=1
+      if (( opt_S && opt_W ))
       then
-        __git_ignore_line __git_changed-in-working-tree_files
-      elif [[ -z ${opt_args[(I)-W|--worktree]} ]]
+        __git_ignore_line __git_changed_files && ret=0
+      elif (( opt_S ))
       then
-        __git_ignore_line __git_changed-in-index_files
+        __git_ignore_line __git_changed-in-index_files && ret=0
       else
-        __git_ignore_line __git_tree_files ${PREFIX:-.} ${(Qv)opt_args[(i)-s|--source]:-HEAD} && ret=0
+        __git_ignore_line __git_changed-in-working-tree_files && ret=0
       fi
     ;;
     sources)
```

danielshahaf:
Thanks for the revision.  I've tested it using `git restore -p` and:

- `git restore <TAB>`, `git restore -S <TAB>`, and `git restore -W <TAB>` offer exactly what they should: no more, no less.

- `git restore -SW <TAB>` offers everything it should offer, and additionally offers files that differ between HEAD and index but don't differ between index and worktree.  This is not a blocker, though: it's better to complete too much than too little.

    Furthermore, `git restore -SWp` actually prompted me for hunks to restore, even though the man page says —

    > When both --staged and --worktree are specified, --source must also be specified

    — which implies that `git restore -SWp` should have been a syntax error.  It seems the man page is out of date?

- However, the patch seems to have dropped support for completing files from _another_ tree after `git restore --source=$tree`, hasn't it?  Could/Shouldn't that be re-added?  There's a `__git_treeish-to-index_files` helper that could be used here (and we can add new helpers if needed).

Cheers!

harogaston:
Great. I hadn't tried `-p` so that is good job. I will have a look at your last point. Thank you! 
I'll get back to you as soon as I can.

danielshahaf:
Looking forward to it. I'd mainly like to endure we don't introduce regressions by accident. Cheers!

Excuse brevity (touch keyboard).

harogaston:
I don't know if I understood correctly what you said about "`git restore --source=$tree` dropping support for completing files from another tree" But if I did the only thing missing was the `=` sign, I have been testing with `--sources <treeish_ref>` so I didn't notice, but the documentation has the `=` sign although it work both ways in practice.

If I got you right then here's the patch. If not please let me know! 
Cheers

danielshahaf:
Thanks for adding that `=` sign.  So long as you're there, should `{-s,…}` be changed to `{-s+,…}` as well?  That will affect the argument to the short option: it'll allow completing `-s<TAB>` without a space, in addition to `-s <TAB>` with a space.

However, that's not quite what I meant with my comment.  The code in master runs `__git_tree_files ${PREFIX:-.} ${(Qv)opt_args[(i)-s|--source]:-HEAD}`, which offers files from the tree that's the argument to the `--source` option: after `git restore -s foo <TAB>` it interprets `foo` as a treeish and completes files in that.  Your patch seems to remove this handling, doesn't it?

Does this amount to regression in completing filenames when `--source` is passed?

To be clear, I am _not_ asking you to add support for `--source` in combination with `-S` and/or `-W`; I am simply asking whether there is a use-case that already works in master, but will stop working if this PR is merged as it is.  If so, it would be nice to fix _that_ use-case, if feasible.

It's possible that the code in master completes the wrong subset of files when `--source` is specified, in which case there's no regression in the first place; or that there _is_ a regression, but the cost of reimplementing support for the (presumably comparatively rare) case that `--source` is passed outweighs the benefit of correctly supporting `-S` and `-W` in the (presumably more common) case that `--source` isn't passed; etc..  I'd like to know where we stand.

FWIW, if I weren't concerned about a potential regression, I'd have merged the patch as it is, notwithstanding that one case where it offers more files than it should.

harogaston:
Sure, I don't want this to be a regression in any form either. 

Just to be sure we are on the same page. You say that the treeish object specified in `--source` should limit the `<pathspec>` to only valid paths existing under that treeish object?

If that is the case I think I can restore that behavior, I just tried. But I don't think that would be the correct behavior, the pathspec is a way to limit the scope of the operation so it should be taken possibly from the actual working tree.

I apologize for taking so much of your time already.

danielshahaf:
> Just to be sure we are on the same page. You say that the treeish object specified in `--source` should limit the `<pathspec>` to only valid paths existing under that treeish object?
> 

Almost.  I am saying that's what the code in master does.  I consider code already committed to be "innocent until proven guilty" — i.e., I presume it correct until and unless that's disproven.

> If that is the case I think I can restore that behavior, I just tried. But I don't think that would be the correct behavior, the pathspec is a way to limit the scope of the operation so it should be taken possibly from the actual working tree.

Well, let's see:

```
% git log --all --name-status --oneline --decorate --graph 
* 7057520 (topic) add topic
| A     topic
| * bce1f62 (HEAD -> master) add master
|/  
|   A   master
* c9c8303 add iota
  A     iota
% git restore -ps topic                                    
diff --git b/master a/master
index e69de29..0000000
deleted file mode 100644
(1/1) Apply deletion to worktree [y,n,q,a,d,?]? n

diff --git b/topic a/topic
new file mode 100644
index 0000000..e69de29

% 
```

A file that's present in the index and worktree but not in the --source tree was prompted for.

A file that's present in the --source tree but neither in the index nor in the worktree wasn't prompted for.

Therefore, it seems you are correct, and files should be completed from the worktree or index, not from the --source tree.  Therefore, there is no regression in the first place, and the patch should be merged as-is.

@okapia, the `__git_tree_files … ${(Qv)opt_args[(i)-s|--source]:-HEAD}` call was added by you; let us know if we missed anything.

> I apologize for taking so much of your time already.

Don't apologize.  Reviewing pull requests is a good way to spend my time.  I hope we'll see more from you :-)

I'll merge the PR… well, I guess I'll go ahead and merge it now, since @okapia had already been Cc'd to this PR.  Oliver, if you notice any bug here, let us know (even if it's after the merge).

danielshahaf:
Merged in f306221674af05422c9fd60410f2ab054e123255 and a6610c4a17637cde780f9088273929071b742ef8.  (I squashed the commits and tweaked the log messages.)  Thanks a lot!

harogaston:
That's great. Nice you poked @okapia in case we missed something.
I'll be at reach if anything goes high wire.

Cheers @danielshahaf

PR #47, opened 2020-01-31
Pass -dead_strip and -bind_at_load on macOS

rex4539:
From `man ld`:
```
-dead_strip
                 Remove functions and data that are unreachable by the entry point or exported symbols.
```
```
-bind_at_load
                 Sets a bit in the mach header of the resulting binary which tells dyld to bind all symbols when the binary is loaded,
                 rather than lazily.
```
Before:
```
% wc -c zsh   
  733104 zsh

% otool -l zsh
            cmd LC_DYLD_INFO_ONLY
        cmdsize 48
     rebase_off 671744
    rebase_size 944
       bind_off 672688
      bind_size 176
  weak_bind_off 0
 weak_bind_size 0
  lazy_bind_off 672864
 lazy_bind_size 2992
     export_off 675856
    export_size 18856
```
After:
```
% wc -c zsh
  732224 zsh

% otool -l zsh
            cmd LC_DYLD_INFO_ONLY
        cmdsize 48
     rebase_off 671744
    rebase_size 944
       bind_off 672688
      bind_size 2288
  weak_bind_off 0
 weak_bind_size 0
  lazy_bind_off 0
 lazy_bind_size 0
     export_off 674976
    export_size 18856
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -489,7 +489,7 @@ fi
 if test -n "$auto_ldflags"; then
   case "${enable_zsh_debug}$host_os" in
     yesaix*|yeshpux*|yesnetbsd*|yesopenbsd*) ;;  # "ld -g" is not valid on these systems
-    darwin*) LDFLAGS=-Wl,-x ;;
+    darwin*) LDFLAGS=-Wl,-x,-dead_strip,-bind_at_load ;;
     yes*)    LDFLAGS=-g ;;
     *)       LDFLAGS=-s ;;
   esac


danielshahaf:
@rex4539 Sorry, but I don't have a macOS build environment and I'm not familiar with these flags.  Judging by the man pages you quote (thanks for doing that, by the way), it's not clear to me whether `-dead_strip` is appropriate since zsh does use dlopen() to load modules and config.modules may be used to link modules statically, and it's not clear to me why `bind_at_load` should on by default in non-`--enable-zsh-debug` builds.  I suggest that you follow up on zsh-workers@.

Thanks for the patch!

rex4539:
By the way, `bind_at_load` is the equivalent of `-Wl,-z,now` on Linux.

The whole idea of this patch is to make the binary slightly lighter.

rex4539:
If it's of any help, all tests pass.

danielshahaf:
I assumed so, but the question is would the tests pass under other build configurations as well, under other versions of the OS and compiler/linker toolchain, etc..

Let's wait for someone else to have a look — I think Oliver and/or dana might be better placed to review this.

okapia:
I tend to think that the default build shouldn't deviate from default compiler options except where necessary for a successful build. autoconf makes it easy to add extra options but removing ones that have been hardcoded can be tricky. Packaging systems often apply their own set of build options. Why doesn't Apple make these options the default? Isn't lazy loading perhaps a useful feature - if zsh links, e.g. pcre, does it not give a saving that the library isn't loaded until the feature is used?

We could perhaps document recommended sets of options in the MACHINES file.

rex4539:
Withdrawing due to no interest.

PR #48, opened 2020-02-16
vcs_info git: avoid warnings in bare repositories

bk2204:
Git 2.25 introduced a change to how git rev-parse --show-toplevel behaves.  Traditionally, it printed an empty line if the user was in a bare repository.  Now it dies, printing an error to standard error. Consequently, when the user is in a bare repository with a newer Git, vcs_info prints noisily to standard error.

While this is functionally harmless, it is annoying for the shell to print messages from Git every time the prompt is printed, so let's silence the error message.

### DIFF
--- a/Functions/VCS_Info/Backends/VCS_INFO_get_data_git
+++ b/Functions/VCS_Info/Backends/VCS_INFO_get_data_git
@@ -139,6 +139,9 @@ VCS_INFO_git_handle_patches () {
 gitdir=${vcs_comm[gitdir]}
 VCS_INFO_git_getbranch ${gitdir}
 gitbase=$( ${vcs_comm[cmd]} rev-parse --show-toplevel 2> /dev/null )
+if [[ -z ${gitbase} ]]; then
+    gitbase=${gitdir:P}
+fi
 rrn=${gitbase:t}
 if zstyle -t ":vcs_info:${vcs}:${usercontext}:${rrn}" get-revision ; then
     gitsha1=$(${vcs_comm[cmd]} rev-parse --quiet --verify HEAD)
--- a/Functions/VCS_Info/Backends/VCS_INFO_get_data_git
+++ b/Functions/VCS_Info/Backends/VCS_INFO_get_data_git
@@ -138,7 +138,7 @@ VCS_INFO_git_handle_patches () {
 
 gitdir=${vcs_comm[gitdir]}
 VCS_INFO_git_getbranch ${gitdir}
-gitbase=$( ${vcs_comm[cmd]} rev-parse --show-toplevel )
+gitbase=$( ${vcs_comm[cmd]} rev-parse --show-toplevel 2> /dev/null )
 rrn=${gitbase:t}
 if zstyle -t ":vcs_info:${vcs}:${usercontext}:${rrn}" get-revision ; then
     gitsha1=$(${vcs_comm[cmd]} rev-parse --quiet --verify HEAD)


danielshahaf:
We should still set `$rrn` to the correct value: it's part of the API.

Does git print to stdout the value older versions print? If not, can we get that information by some other means?

danielshahaf:
Sorry, I missed the part where you said git used to print an empty line there.  (Incidentally, my system `--show-toplevel` in a bare repository exits 0 with no output whatsoever, not even an empty line.)  However, I still thinks it would be better to set `$rrn` to the basename of the repository root for bare repositories, since the API promises that.  Is that doable?

bk2204:
Sorry, I may have misremembered and it did print nothing at all.  It is possible to set `$rrn` to the Git directory if it's a bare repository, with something like the following:

```sh
gitbase=$( ${vcs_comm[cmd]} rev-parse --show-toplevel 2> /dev/null )
rrn=${gitbase:=$gitdir:t}
```

Is that the direction you'd like to go?

danielshahaf:
> Is that the direction you'd like to go?
> 

Yes, please.  $rrn is an API both in its own right (documented in
Misc/vcs_info-examples) and affects the zstyle contexts.  When
operating on a bare repository I would expect $rrn to be the basename
of the git dir, as you said.

bk2204:
I've pushed up an additional commit which does that, since I think detecting bare repositories is a logically separate (but desirable) change.

Functions/VCS_Info/Backends/VCS_INFO_get_data_git
```diff
@@ -138,8 +138,8 @@ VCS_INFO_git_handle_patches () {
 
 gitdir=${vcs_comm[gitdir]}
 VCS_INFO_git_getbranch ${gitdir}
-gitbase=$( ${vcs_comm[cmd]} rev-parse --show-toplevel )
-rrn=${gitbase:t}
+gitbase=$( ${vcs_comm[cmd]} rev-parse --show-toplevel 2> /dev/null )
+rrn=${gitbase:=$gitdir:t}
```
danielshahaf: You lost the `:t` modifier in the non-bare repository case.

Could you avoid assigning to `gitbase` on the line that assigns to $rrn, for readability?

```suggestion
if [[ -z $gitbase ]]; then
    gitbase=$gitdir # bare repository
fi
rrn=${gitbase:t}
```

bk2204:
Fixed.

danielshahaf:
Sorry, still not quite right:

```zsh
% git init foo
Initialized empty Git repository in /scratch/tmp.P4vYKhiwWG/foo/.git/
% git init --bare bar 
Initialized empty Git repository in /scratch/tmp.P4vYKhiwWG/bar/
% +vi-f() typeset -p rrn ; vcs_info_hookadd post-backend f 
% cd foo
typeset -g -r rrn=foo                                                                                                                                                                                               
% cd ../bar
typeset -g -r rrn=.                                                                                                                                                                                                 
% cd
```

The second case should say `bar`.  I guess a `:P` needs to be added somewhere.

bk2204:
Ah, yes, you're using `--git-dir`, not `--absolute-git-dir`, so `:P` is needed.  Should be fixed now.

danielshahaf:
Thanks, LGTM. I'll add a comment and merge.

Is there any reason to call `--is-bare-repository` explicitly? (Not requesting a revision; just asking.)

danielshahaf:
Merged in 51260963703bbc91944434648ae71a05addbcb05.  I added the comment and, since I didn't want to delay further, also took the liberty of editing the first log message to remove the statement about printing an empty line, per the above.  Thanks for the patch!

bk2204:
> Is there any reason to call `--is-bare-repository` explicitly? (Not requesting a revision; just asking.)

No, I don't think so.  I think what we have is fine and we can determine whether it's bare from what we have if we need to in the future.  This will result in fewer git invocations which should make the prompt a little faster.

PR #49, opened 2020-03-05
typo: longson should be loongson

linkslice:


Completion/BSD/Type/_obsd_architectures
```diff
@@ -3,4 +3,4 @@
 local expl
 
 _description architectures expl 'architecture'
-compadd "$@" "$expl[@]" alpha amd64 arm64 armv7 hppa i386 landisk longson luna88k macppc octeon sgi sparc64
+compadd "$@" "$expl[@]" alpha amd64 arm64 armv7 hppa i386 landisk loongson luna88k macppc octeon sgi sparc64
```
danielshahaf: Should `sgi` be retained or removed? I don't see it under https://github.com/zsh-users/zsh/blob/243083c50b7cb8f7b1539abcf8418ebb3163865e/Completion/BSD/Type/_obsd_architectures#L6.

danielshahaf: ... under https://www.openbsd.org/faq/faq4.html#Download.

danielshahaf:
Merged in 2f419b332d9cf28eaad0523b3404ab8ce52a82e0 with a tweaked log message.  Let us know about `sgi`.  Thanks for the patch!

Completion/BSD/Type/_obsd_architectures
linkslice: probably not. Support for was dropped with the latest release. https://www.openbsd.org/sgi.html

danielshahaf: Thanks, done in dd50f125b5eb65896642d2ff664adefd33f1004c.

PR #50, opened 2020-03-08
_ssh_hosts: Complete hosts in "Match" directive

fumiyas:
Support SSH hostname completion in "Match host ..." and so on.

```
Match host foo
...
Match originalhost bar
...
Match final baz
...
Match canonical qux
...
```

### DIFF
--- a/Completion/Unix/Type/_ssh_hosts
+++ b/Completion/Unix/Type/_ssh_hosts
@@ -19,8 +19,20 @@ fi
 if [[ -r $config ]]; then
   local key line host
   local -a lines=("${(@f)$(<"$config")}") 2>/dev/null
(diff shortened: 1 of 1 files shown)


fumiyas:
Ping. Should I send the patch to the zsh-workers mailing list?

danielshahaf:
Hi, sorry for the radio silence.  I've just sent a ping to -workers@ for you, covering a few other PRs as well.  I took the liberty of Bcc'ing you.  (Bcc, rather than Cc, so you wouldn't be spammed by unrelated responses.)

okapia:
Looks good to me and Bart seemed happy in the -workers thread.
Pushed in a0a2747
Thanks for the contribution.

PR #53, opened 2020-04-20
completion: Allow specifying directory name for _remote_files

insidewhy:
I use this like this:

```zsh
function get_media_file() {
  rsync -aP media-server:media-files/$1 .
}
function _get_media_file() {
  _remote_files -h media-server -W media-files -- ssh
}
compdef _get_media_file get_media_file
```

EDIT: Updated the example to switch `-d` to `-W` as per the code review comment.

### DIFF
--- a/Completion/Unix/Type/_remote_files
+++ b/Completion/Unix/Type/_remote_files
@@ -4,13 +4,16 @@
 # key-based authentication with no passwords or a running ssh-agent to work.
 #
 # Usage:
-#   _remote_files [-/] [-g glob] [-h host] -- <cmd> [<cmd options>]
+#   _remote_files [-/] [-g glob] [-h host] [-W dir] -- <cmd> [<cmd options>]
 #
 # Options:
 # - -/: only complete directories
 # - -g: specify a pattern to match against files
 #       p, = and * glob qualifiers supported
 # - -h: specify the remote host, default is ${IPREFIX%:}
+# - -W: specify the parent directory to list files from,
+#       default is the home directory
+#
 #
 # Commands:
 # - ssh: Additional options for non-interactive use are automatically added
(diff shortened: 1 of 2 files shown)


Completion/Unix/Type/_remote_files
```diff
@@ -30,12 +30,12 @@
...
 local expl rempat remfiles remdispf remdispd args cmd suf ret=1
 local -a args cmd_args
-local glob host
+local glob host dir dirprefix
 
 if zstyle -T ":completion:${curcontext}:files" remote-access; then
 
   # Parse options to _remote_files. Stops at the first "--".
-  zparseopts -D -E -a args / g:=glob h:=host
+  zparseopts -D -E -a args / g:=glob h:=host d:=dir
```
danielshahaf, line 38: Could you add the new flag to the documentation (lines 6 through 13)?

insidewhy: @danielshahaf Thanks for your review, I've added the docs.

okapia: This change looks fine to me, the one thing I wonder is whether the switch should be `-W` for consistency with `_files`. `-d` is a compadd option, albeit one that makes little sense here but they tend to be best avoided. I don't think it necessarily needs to replicate the `_files` support for an array with `-W` – that doesn't get much use anyway,

insidewhy: @okapia Thanks for your help, I've updated the code to switch `-d` with `-W`.

okapia:
Thanks for the contribution and for addressing the comments promptly. I've merged this in dc8061e.
We should probably add `_remote_files` to the main yodl documentation as it seems to get wider use than just `_rsh` and `_ssh` as originally but that's a separate issue.

PR #54, opened 2020-04-21
Fix dig completion error

dhath:
Two `compadd` commands in `_dig` were changed in 2018, fixing one (for `-c`) but breaking the other.  They have the same arguments, but the arguments have different escape requirements.  In this case, the `:` should not be escaped.

### DIFF
--- a/Completion/Unix/Command/_dig
+++ b/Completion/Unix/Command/_dig
@@ -89,7 +89,7 @@ if [[ -n $state ]]; then
     _wanted hosts expl 'DNS server' _hosts && ret=0;
   else
     case $#line in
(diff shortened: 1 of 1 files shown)


danielshahaf:
Thanks for the patch.

@okdana, could you have a look? The original change was yours.

okdana:
Looks like a copy/paste error on my part, sorry. Merged the fix, ty

danielshahaf:
Thanks :)

@zsh-users/Administrators Would you please give dana whatever permissions she needs in order to be able to close issues in this repository, seeing as she has commit access upstream?  I guess that's membership of the https://github.com/orgs/zsh-users/teams/zsh-committers team.

PR #55, opened 2020-04-26
git-add completion: offer files off the bat, not one path component at a time

amynbe:
Before:
before
After:
after
Initially proposed on reddit:
https://www.reddit.com/r/zsh/comments/ass2tc/gitadd_completion_with_full_paths_listed_at_once/

### DIFF
--- a/Completion/Unix/Command/_git
+++ b/Completion/Unix/Command/_git
@@ -7130,12 +7130,27 @@ __git_deleted_files () {
 
 (( $+functions[__git_modified_files] )) ||
 __git_modified_files () {
-  __git_files --modified modified-files 'modified file' $*
+  __git_status_files
+}
+
+(( $+functions[__git_status_files] )) ||
+__git_status_files () {
+  local -a status_files=( ${"${(0)"$(git status -z)"}"} )
+
+  local -a unstaged_files_with_status=( ${(M)status_files:#?[^ ]*} )
+  local -a unstaged_files=( ${unstaged_files_with_status/?? /} )
+  _describe -t unstaged 'Unstaged' unstaged_files && ret=0
+
+  local -a staged_files_with_status=( ${(M)status_files:#[^ \?]*} )
+  local -a staged_files=( ${staged_files_with_status/?? /} )
+  _describe -t staged 'Staged' staged_files && ret=0
+
+  return $ret
 }
 
 (( $+functions[__git_other_files] )) ||
 __git_other_files () {
-  __git_files --others untracked-files 'untracked file' $*
+
 }
 
 (( $+functions[__git_ignored_cached_files] )) ||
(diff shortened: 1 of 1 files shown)


Completion/Unix/Command/_git
```diff
@@ -7130,12 +7130,27 @@ __git_deleted_files () {
...
+  local -a staged_files=( ${staged_files_with_status/?? /} )
+  _describe -t staged 'Staged' staged_files && ret=0
+
+  return $ret
 }
 
 (( $+functions[__git_other_files] )) ||
 __git_other_files () {
-  __git_files --others untracked-files 'untracked file' $*
+
```
danielshahaf, line 7153: 1. Same issue with `$*` as above.
2. Leaving the function empty would break `_git-clean`, which also calls this function.

Completion/Unix/Command/_git
```diff
@@ -7179,8 +7194,14 @@ __git_changed-in-index_files () {
 
 (( $+functions[__git_treeish-to-index_files] )) ||
 __git_treeish-to-index_files () {
-  local tree=$1; shift
-  __git_diff-index_files $tree "files different between ${(qq)tree} and the index" treeish-to-index-files "$@"
+  __git_staged_files
+}
```
danielshahaf, line 7198: Given that the incumbent code uses `$1` and the new code doesn't, I don't see how this can possibly be correct.

Completion/Unix/Command/_git
```diff
@@ -7130,12 +7130,27 @@ __git_deleted_files () {
 
 (( $+functions[__git_modified_files] )) ||
 __git_modified_files () {
-  __git_files --modified modified-files 'modified file' $*
+  __git_status_files
+}
```
danielshahaf, line 7134: Given that the incumbent code uses `$*` and the new code doesn't, I don't see how this can possibly be correct.  For starters, I expect it breaks completion of ignored files after `git add -f <TAB>` (see `_git-add`).

amynbe:
I'm just getting back to this snippet after writing it a long time ago and thought it was better contributed upstream, so apologies for the random PR.

I think the problem reveals itself when you mix untracked and modified files under the same subfolder:

```
% cd "$(mktemp -d)"
% git init
% touch foo
% mkdir src
% touch src/bar1 src/bar2
% git add src/bar1 && git commit -madd
% echo >>src/bar1
% git add <TAB>
 -- modified file --
src
 -- untracked file --
foo  src
```

You can see that `src` appears twice with no way of selecting a specific file inside it.

With my change:

```
% git add <TAB>
 -- Unstaged --
foo       src/bar1  src/bar2
```

danielshahaf:
> You can see that `src` appears twice with no way of selecting a specific file inside it.

Once you select `src` you'll be able to complete modified/untracked files inside it.  You can even do `git add s/b<TAB>` to complete files directly.  This functionality is provided by the `_multi_parts` call down in `__git_files`.

I see that offering `src/bar1` and `src/bar2` right off the bat can be useful.  However, I'm not sure how best to implement it.  I wonder if it would make sense to teach `_multi_parts` some sort of `dont-try-to-be-smart-just-offer-all-matches-right-off-the-bat` zstyle.  This would solve the problem for _every_ use of `_multi_parts`.

Could you follow up on the mailing list?  There are more eyeballs there.

danielshahaf:
Retitled.

danielshahaf:
> Could you follow up on the mailing list? There are more eyeballs there.

I see you've done that (thanks), but for posterity, here's the thread: [workers/45721](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=45721).

PR #57, opened 2020-05-03
Support an 'owner' key in region_highlight.

danielshahaf:
This is to allow plugins to easily remove only entries they had added.  See https://github.com/zsh-users/zsh-syntax-highlighting/issues/418 and the cross-referenced issues.

### DIFF
--- a/Doc/Zsh/zle.yo
+++ b/Doc/Zsh/zle.yo
@@ -1014,13 +1014,13 @@ any predisplay string should be highlighted in bold.
 Note that the effect of tt(region_highlight) is not saved and disappears
 as soon as the line is accepted.
 
-Note that zsh 5.8 and older does not support the `tt(memo=)var(token)' field
+Note that zsh 5.8 and older do not support the `tt(memo=)var(token)' field
 and may misparse the third (highlight specification) field when a memo
 is given.
 COMMENT(The syntax `tt(0 20 bold, memo=foobar)' (with an auxiliary comma)
-happens to work on both zsh 5.8 and zsh 5.9-to-be, but I'm deliberately leaving
-that undocumented as it's more of an accident of implementation than something
-I'm comfortable making a first-class-citizen API promise.)
+happens to work on both zsh <=5.8 and zsh 5.9-to-be, but that seems to be more of
+an accident of implementation than something we should make a first-class-citizen
+API promise.  It's mentioned in the "Incompatibilities" section of README.)
 
 The final highlighting on the command line depends on both tt(region_highlight)
 and tt(zle_highlight); see
--- a/Doc/Zsh/zle.yo
+++ b/Doc/Zsh/zle.yo
@@ -1008,6 +1008,14 @@ any predisplay string should be highlighted in bold.
 Note that the effect of tt(region_highlight) is not saved and disappears
 as soon as the line is accepted.
 
+Note that zsh 5.8 and older does not support the `tt(memo=)var(token)' field
+and may misparse the third (highlight specification) field when a memo
+is given.
+COMMENT(The syntax `tt(0 20 bold, memo=foobar)' (with an auxiliary comma)
+happens to work on both zsh 5.8 and zsh 5.9-to-be, but I'm deliberately leaving
+that undocumented as it's more of an accident of implementation than something
+I'm comfortable making a first-class-citizen API promise.)
+
 The final highlighting on the command line depends on both tt(region_highlight)
 and tt(zle_highlight); see
 ifzman(the section CHARACTER HIGHLIGHTING below)\
--- a/Doc/Zsh/zle.yo
+++ b/Doc/Zsh/zle.yo
@@ -989,15 +989,18 @@ ifnzman(noderef(Character Highlighting))\
 ifzman(the section `Character Highlighting' below);
 for example, tt(standout) or tt(fg=red,bold).)
 itemiz(Optionally, a string of the form `tt(memo=)var(token)'.
-The var(token) value is preserved verbatim but not parsed in any way.
-Plugins may use this to identify elements they had added, for example, in order
-to remove them with
-`tt(region_highlight=+LPAR() ${region_highlight:#*memo=)var(token)tt(} +RPAR())'.)
+The var(token) consists of everything between the `tt(=)' and the next
+whitespace, comma, NUL, or the end of the string.
+The var(token) is preserved verbatim but not parsed in any way.
+Plugins may use this to identify array elements they have added: for example,
+a plugin might set var(token) to its (the plugin's) name and then use
+`tt(region_highlight=+LPAR() ${region_highlight:#*memo=)var(token)tt(} +RPAR())'
+in order to remove array elements it have added.)
 enditemize()
 
 For example, 
 
-example(region_highlight=("P0 20 bold memo=myplugin"))
+example(region_highlight=("P0 20 bold memo=foobar"))
 
 specifies that the first twenty characters of the text including
 any predisplay string should be highlighted in bold.
--- a/Doc/Zsh/zle.yo
+++ b/Doc/Zsh/zle.yo
@@ -988,16 +988,16 @@ used for contexts in the parameter tt(zle_highlight), see
 ifnzman(noderef(Character Highlighting))\
 ifzman(the section `Character Highlighting' below);
 for example, tt(standout) or tt(fg=red,bold).)
-itemiz(Optionally, a string of the form `tt(owner=)var(token)'.
+itemiz(Optionally, a string of the form `tt(memo=)var(token)'.
 The var(token) value is preserved verbatim but not parsed in any way.
 Plugins may use this to identify elements they had added, for example, in order
 to remove them with
-`tt(region_highlight=+LPAR() ${region_highlight:#*owner=)var(token)tt(} +RPAR())'.)
+`tt(region_highlight=+LPAR() ${region_highlight:#*memo=)var(token)tt(} +RPAR())'.)
 enditemize()
 
 For example, 
 
-example(region_highlight=("P0 20 bold owner=myplugin"))
+example(region_highlight=("P0 20 bold memo=myplugin"))
 
 specifies that the first twenty characters of the text including
 any predisplay string should be highlighted in bold.
--- a/Doc/Zsh/zle.yo
+++ b/Doc/Zsh/zle.yo
@@ -992,10 +992,16 @@ itemiz(Optionally, a string of the form `tt(memo=)var(token)'.
 The var(token) consists of everything between the `tt(=)' and the next
 whitespace, comma, NUL, or the end of the string.
 The var(token) is preserved verbatim but not parsed in any way.
+
 Plugins may use this to identify array elements they have added: for example,
 a plugin might set var(token) to its (the plugin's) name and then use
 `tt(region_highlight=+LPAR() ${region_highlight:#*memo=)var(token)tt(} +RPAR())'
-in order to remove array elements it have added.)
+in order to remove array elements it have added.
+
+(This example uses the `tt(${)var(name)tt(:#)var(pattern)tt(})' array-grepping
+syntax described in
+ifzman(the section `Parameter Expansion' in zmanref(zshexpn))\
+ifnzman(noderef(Parameter Expansion)).))
 enditemize()
 
 For example, 
--- a/Doc/Zsh/zle.yo
+++ b/Doc/Zsh/zle.yo
@@ -974,27 +974,30 @@ of the non-editable parts of the command line in tt(PREDISPLAY)
 and tt(POSTDISPLAY) are possible, but note that the tt(P) flag
 is needed for character indexing to include tt(PREDISPLAY).
 
-Each string consists of the following parts:
+Each string consists of the following whitespace-separated parts:
 
 startitemize()
 itemiz(Optionally, a `tt(P)' to signify that the start and end offset that
 follow include any string set by the tt(PREDISPLAY) special parameter;
 this is needed if the predisplay string itself is to be highlighted.
-Whitespace may follow the `tt(P)'.)
-itemiz(A start offset in the same units as tt(CURSOR), terminated by
-whitespace.)
-itemiz(An end offset in the same units as tt(CURSOR), terminated by
-whitespace.)
+Whitespace between the `tt(P)' and the start offset is optional.)
+itemiz(A start offset in the same units as tt(CURSOR).)
+itemiz(An end offset in the same units as tt(CURSOR).)
 itemiz(A highlight specification in the same format as
 used for contexts in the parameter tt(zle_highlight), see
 ifnzman(noderef(Character Highlighting))\
 ifzman(the section `Character Highlighting' below);
-for example, tt(standout) or tt(fg=red,bold)).
+for example, tt(standout) or tt(fg=red,bold).)
+itemiz(Optionally, a string of the form `tt(owner=)var(token)'.
+The var(token) value is preserved verbatim but not parsed in any way.
+Plugins may use this to identify elements they had added, for example, in order
+to remove them with
+`tt(region_highlight=+LPAR() ${region_highlight:#*owner=)var(token)tt(} +RPAR())'.)
 enditemize()
 
 For example, 
 
-example(region_highlight=("P0 20 bold"))
+example(region_highlight=("P0 20 bold owner=myplugin"))
 
 specifies that the first twenty characters of the text including
 any predisplay string should be highlighted in bold.
--- a/Src/Zle/zle_refresh.c
+++ b/Src/Zle/zle_refresh.c
@@ -519,7 +519,7 @@ set_region_highlight(UNUSED(Param pm), char **aval)
 
 	if (strpfx(memo_equals, strp)) {
 	    const char *memo_start = strp + strlen(memo_equals);
-	    const char *memo_end = memo_start;
+	    const char *i, *memo_end;
 
 	    /* 
 	     * Forward compatibility: end parsing at a comma or whitespace to
@@ -527,14 +527,26 @@ set_region_highlight(UNUSED(Param pm), char **aval)
 	     *
 	     * - A fifth field: "0 20 bold memo=foo bar".
 	     *
-	     * - Additional attributes in the fourth field: "0 20 bold memo=foo,bar".
+	     * - Additional attributes in the fourth field: "0 20 bold memo=foo,bar"
+	     *   and "0 20 bold memo=foo\0bar".
 	     *
 	     * For similar reasons, we don't flag an error if the fourth field
 	     * doesn't start with "memo=" as we expect.
 	     */
-	    while (*memo_end && *memo_end != ',' && !inblank(*memo_end))
-		++memo_end;
+	    i = memo_start;
 
+	    /* ### TODO: Consider optimizing the common case that memo_start to
+	     *           end-of-string is entirely ASCII */
+	    while (1) {
+		int nbytes;
+		convchar_t c = unmeta_one(i, &nbytes);
+
+		if (c == '\0' || c == ',' || inblank(c)) {
+		    memo_end = i;
+		    break;
+		} else
+		    i += nbytes;
+	    }
 	    rhp->memo = ztrduppfx(memo_start, memo_end - memo_start);
 	} else
 	    rhp->memo = NULL;
--- a/Src/Zle/zle_refresh.c
+++ b/Src/Zle/zle_refresh.c
@@ -485,6 +485,7 @@ set_region_highlight(UNUSED(Param pm), char **aval)
 	 *aval;
 	 rhp++, aval++) {
 	char *strp, *oldstrp;
+	const char memo_equals[] = "memo=";
 
 	oldstrp = *aval;
 	if (*oldstrp == 'P') {
@@ -516,8 +517,8 @@ set_region_highlight(UNUSED(Param pm), char **aval)
 	while (inblank(*strp))
 	    strp++;
 
-	if (strpfx("memo=", strp))
-	    rhp->memo = ztrdup(strp + 5);
+	if (strpfx(memo_equals, strp))
+	    rhp->memo = ztrdup(strp + strlen(memo_equals));
 	else
 	    rhp->memo = NULL;
     }
--- a/Src/Zle/zle_refresh.c
+++ b/Src/Zle/zle_refresh.c
@@ -212,9 +212,9 @@ static zattr default_atr_on, special_atr_on;
 
 /*
  * Array of region highlights, no special termination.
- * The first element (0) always describes the region between
- * point and mark.  Any other elements are set by the user
- * via the parameter region_highlight.
+ * The first N_SPECIAL_HIGHLIGHTS elements describe special uses of
+ * highlighting, documented under N_SPECIAL_HIGHLIGHTS.
+ * Any other elements are set by the user via the parameter region_highlight.
  */
 
 /**/
@@ -414,16 +414,19 @@ get_region_highlight(UNUSED(Param pm))
 	 arrsize--;
 	 rhp++, arrp++) {
 	char digbuf1[DIGBUFSIZE], digbuf2[DIGBUFSIZE];
-	int atrlen = 0, alloclen;
+	int atrlen, alloclen;
+	const char owner_equals[] = "owner=";
 
 	sprintf(digbuf1, "%d", rhp->start);
 	sprintf(digbuf2, "%d", rhp->end);
 
 	atrlen = output_highlight(rhp->atr, NULL);
 	alloclen = atrlen + strlen(digbuf1) + strlen(digbuf2) +
-	    3; /* 2 spaces, 1 0 */
+	    3; /* 2 spaces, 1 terminating NUL */
 	if (rhp->flags & ZRH_PREDISPLAY)
 	    alloclen += 2; /* "P " */
+	if (rhp->owner)
+	    alloclen += 1 /* space */ + strlen(owner_equals) + strlen(rhp->owner);
 	*arrp = (char *)zhalloc(alloclen * sizeof(char));
 	/*
 	 * On input we allow a space after the flags.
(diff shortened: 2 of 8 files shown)


Doc/Zsh/zle.yo
```diff
@@ -974,27 +974,30 @@ of the non-editable parts of the command line in tt(PREDISPLAY)
...
+The var(token) value is preserved verbatim but not parsed in any way.
+Plugins may use this to identify elements they had added, for example, in order
+to remove them with
+`tt(region_highlight=+LPAR() ${region_highlight:#*owner=)var(token)tt(} +RPAR())'.)
 enditemize()
 
 For example, 
 
-example(region_highlight=("P0 20 bold"))
+example(region_highlight=("P0 20 bold owner=myplugin"))
```
danielshahaf, line 1000: Older zsh's parse this as equivalent to `P0 20 none` unless a comma is added after `bold`.  Is there a better option?

Src/Zle/zle_refresh.c
```diff
@@ -502,7 +511,15 @@ set_region_highlight(UNUSED(Param pm), char **aval)
...
 	    strp++;
 
-	match_highlight(strp, &rhp->atr);
+	strp = (char*) match_highlight(strp, &rhp->atr);
+
+	while (inblank(*strp))
+	    strp++;
+
+	if (strpfx("owner=", strp))
+	    rhp->owner = ztrdup(strp + 6);
```
danielshahaf, line 520: Given the above comment, this shouldn't copy the entire rest of the string, but only up to the next space, for extensibility (forward compatibility).

Doc/Zsh/zle.yo
okapia: Hold on, doesn't using a comma separator give us backward compatibility?
`region_highlight=("P0 20 bold,owner=myplugin"` seems to work in current zsh.

danielshahaf: Thanks for the review.

I considered using a comma separator, but (1) the owner's identity seemed like metadata, not an intrinsic part of the typographic specification; and (2) the syntax in the PR was mildly easier to implement.

I'll sleep on a numeric `precedence=` key (à la add-zle-hook-widget style values) and solving the region/standout duplication issue and get back to you on them.

okapia: Fair enough. Do I understand correctly that the only downside of your approach (with a space) is that for backwards compatibility, you need a trailing comma on the attributes if there isn't one, e.g. `bold,` I agree that it logically belongs separately so is cleaner.

Any precedence feature is a separate matter really. It just came to mind when looking at this. Rather than a numeric `precedence=` key, I was just thinking of an array listing owners in order. By default it'd just contain an entry for the builtin stuff. Plugins would probably add themself to either the beginning or end and users could always reorder them. There are probably other schemes that could work.

danielshahaf: > I understand correctly that the only downside of your approach (with a space) is that for backwards compatibility, you need a trailing comma on the attributes if there isn't one, e.g. `bold,`

It's the only downside I'm aware of, yes.  But there's a reason I've posted this for review :)

> Any precedence feature is a separate matter really. It just came to mind when looking at this. Rather than a numeric `precedence=` key, I was just thinking of an array listing owners in order. By default it'd just contain an entry for the builtin stuff. Plugins would probably add themself to either the beginning or end and users could always reorder them. There are probably other schemes that could work.

The idea being that plugin authors are better placed than users to determine the best relative order of highlights, I guess?  For example, if zsh-autosuggestions were to highlight a suggestion that had just been inserted, that would probably want to apply after z-sy-h's highlights?

Compare the proof-of-concept docs patch I posted in https://github.com/zsh-users/zsh-autosuggestions/issues/529#issue-612034448 under the heading "Describe alternatives you've considered".  (None of the dicsussion in the issue is about that diff, though the LASTWIDGET issue in Eric's latest comment may be of independent interest.)

ericfreese:
Just brainstorming a bit.

As a plugin author, setting an "owner" value is a bit more broad of a use case than I need. All I really need is to be able to uniquely identify certain highlights that I created so that I can delete them later on. If when I create a new highlight, I could get back an automatically generated unique identifier for that highlight in REPLY, that would be enough for me.

@danielshahaf what would you think of adding a new flag (maybe `-h`?) to the `zle` built-in? It would save a specified highlight and give back a unique identifier for that highlight in `REPLY`. The unique identifier could then be provided to remove the highlight.

```
local REPLY

# Add a new highlight with `zle -h`
zle -h "P0 20 bold"

# Remove the highlight by adding `-d` flag
zle -h -d $REPLY
```

I'm thinking that highlights added this way would not be accessible through the `region_highlight` variable and vice versa, but highlights from both would be applied when rendering.

danielshahaf:
> Just brainstorming a bit.
> 

Thanks for the review.

> As a plugin author, setting an "owner" value is a bit more broad of a use case than I need. All I really need is to be able to uniquely identify certain highlights that I created so that I can delete them later on. If when I create a new highlight, I could get back an automatically generated unique identifier for that highlight in REPLY, that would be enough for me.
> 

Could you elaborate on why z-asug needs _unique_ identifiers?  I take it z-asug's use-case wouldn't be addressed by a hardcoding a fixed string or three, but why?

In terms of the PR as it stands, the closest equivalent to your functionality would be something along the lines of:

```zsh
owner="z-asug:$RANDOM:"
region_highlight+=( "P0 20 bold owner=$owner" )
```

and afterwards remove it using the method described in the docs change in this PR.

(Aside: I'll change "owner" in the PR to some more generic term before posting/merging this.)

> @danielshahaf what would you think of adding a new flag (maybe `-h`?) to the `zle` built-in? It would save a specified highlight and give back a unique identifier for that highlight in `REPLY`. The unique identifier could then be provided to remove the highlight.
> 
> ```
> local REPLY
> 
> # Add a new highlight with `zle -h`
> zle -h "P0 20 bold"
> 
> # Remove the highlight by adding `-d` flag
> zle -h -d $REPLY
> ```
> 
> I'm thinking that highlights added this way would not be accessible through the `region_highlight` variable and vice versa, but highlights from both would be applied when rendering.

On the one hand, encapsulating the details _is_ appealing.  I suspect, however, that a blackbox that only supports "add a highlight" and "delete a highlight given its nonce" would be a bit _too_ specialized.

On the other hand, the approach I described above seems to me to be more closely aligned with [PEP 20](https://www.python.org/dev/peps/pep-0020/)'s principles and to be more generic/flexible.  Indeed, it would be easy to implement your proposed API around the API in the PR:

```zsh
zle-h() {
    if [[ $1 == -d ]]; then
        region_highlight=( ${region_highlight:#*owner="$2"} )
    else
        setopt localoptions nowarncreateglobal
        REPLY="$RANDOM.$RANDOM"
        region_highlight+=( "$1 owner=$REPLY" )
    fi
}
```

… whereas the converse would require callers to maintain their own records of which nonces (returned values of $REPLY) identify which highlights.  For example, z-sy-h would have to maintain a simple array of all nonces it has set, and then delete all of them (perhaps using a `zle -h -d "${nonces[@]}"` syntax).  A plugin that needs to distinguish among its own highlights — z-sy-h doesn't — would have to maintain multiple arrays, or an associative array, etc., to track its nonces.

Not sure which I prefer, yet.

ericfreese:
> Could you elaborate on why z-asug needs unique identifiers? I take it z-asug's use-case wouldn't be addressed by a hardcoding a fixed string or three, but why?

The need for unique identifiers is just so that I can be sure that I’m not removing anything else when I delete my highlight. If it weren’t unique, there may be a chance that I remove something else unintentionally. Hard coding a fixed string would work practically but my point was that it’s not functionality that I need. I don’t care what the identifier value is, so having to specify some value for it is just extra work for me to do. It would be nice if zsh could handle it for me. It’s not a big deal, just a thought while we’re coming up with new functionality 😄 

> I'm thinking that highlights added this way would not be accessible through the region_highlight variable and vice versa, but highlights from both would be applied when rendering.

As I think about this more I’m wondering if would be feasible to just store the highlights added via the new zle flag to region_highlight. The unique identifier would then be just an optional attribute of a highlight- present for those added via zle and not present for those added through setting the param.

danielshahaf:
> > Could you elaborate on why z-asug needs unique identifiers? I take it z-asug's use-case wouldn't be addressed by a hardcoding a fixed string or three, but why?
> 
> The need for unique identifiers is just so that I can be sure that I’m 
> not removing anything else when I delete my highlight. If it weren’t 
> unique, there may be a chance that I remove something else 
> unintentionally. Hard coding a fixed string would work practically but 
> my point was that it’s not functionality that I need. I don’t care what 
> the identifier value is, so having to specify some value for it is just 
> extra work for me to do. It would be nice if zsh could handle it for 
> me. It’s not a big deal, just a thought while we’re coming up with new 
> functionality 😄
> 

WDYT of, say, «region_highlight+=( "P0 20 bold autoremove=yes" )» to add
an entry that would be automatically removed (by add-zle-hook-widget or
possibly zle itself) at some appropriate time?  (What would an
appropriate time be in your use-case?)

I'll sleep on the rest of your comment and reply further later.

danielshahaf:
> As I think about this more I’m wondering if would be feasible to just store the highlights added via the new zle flag to region_highlight. The unique identifier would then be just an optional attribute of a highlight- present for those added via zle and not present for those added through setting the param.

Okay, so you're proposing to encapsulate adding a single entry and removing a single entry.  However, operations such as "remove all my entries", "remove all entries", "enumerate entries", etc., would not be encapsulated, so they'd only be possible to implement by directly accessing the array.  Doesn't that somewhat negate the benefits of encapsulation, at least at the `zle` level?  An encapsulation, in this case, would seem to be another API entry point to maintain, but without the benefit of being able to change the implementation without changing the API callers?

ericfreese:
> WDYT of, say, «region_highlight+=( "P0 20 bold autoremove=yes" )» to add
an entry that would be automatically removed (by add-zle-hook-widget or
possibly zle itself) at some appropriate time?  (What would an
appropriate time be in your use-case?)

Hmm yeah this could probably work. The appropriate time for z-asug could probably be any time between the end of a redraw and the start of the next redraw cycle (just before the line-pre-redraw hooks are called). This would probably be the right time for z-sy-h as well. So setting that flag would be kind of like saying "I want this highlight to be around for the next redraw cycle, but remove it after that".

I'm a little apprehensive to give up control of when exactly the entries are removed, but I'm struggling to come up with any real justification for that apprehension. Would there be cases where someone would want a highlight to stick around until they manually remove it?

> Okay, so you're proposing to encapsulate adding a single entry and removing a single entry. However, operations such as "remove all my entries", "remove all entries", "enumerate entries", etc., would not be encapsulated, so they'd only be possible to implement by directly accessing the array.

That's basically what I'm thinking. You could still support "remove all my entries" by passing multiple tokens to `zle -h -d`. And if there's some way to return multiple entries in REPLY, we could add multiple entries using `zle -h` as well. "remove all entries" could be supported by allowing a pattern arg to `zle -h -d` and passing something like `zle -h -d *`. "enumerate entries" could be `zle -h -l` or something. So I think we could wrap the whole thing up in the `zle` interface, but it will be a bit more clunky for those use cases where the param makes more sense.

> Doesn't that somewhat negate the benefits of encapsulation, at least at the zle level?

Somewhat yeah. The benefit we'd get from adding `zle -h` is being able to have side effects when a highlight is added (setting REPLY to a unique identifier for that highlight). I'm thinking we keep `region_highlight` around in its current form mainly for backward compatibility, but there are some other options:

1. Deprecate `region_highlight` and encapsulate highlight behavior entirely in new `zle -h`.
2. Make `region_highlight` read-only and encapsulate adding/removing highlights in `zle -h`. (Isn't this kind of like existing params like `$widgets` that are modified by builtins but available as read-only params for convenience?)

> An encapsulation, in this case, would seem to be another API entry point to maintain, but without the benefit of being able to change the implementation without changing the API callers?

I think the implementation could technically change as long as the API contracts are kept intact. The API contracts being the details of both `zle -h` builtin and `region_highlight` param. For example, the algorithm to generate the unique token for each new  `zle -h` highlight could be altered without any changes required to clients.

danielshahaf:
> > WDYT of, say, «region_highlight+=( "P0 20 bold autoremove=yes" )» to add an entry that would be automatically removed (by add-zle-hook-widget or possibly zle itself) at some appropriate time?  (What would an appropriate time be in your use-case?)
> 
> Hmm yeah this could probably work. The appropriate time for z-asug could probably be any time between the end of a redraw and the start of the next redraw cycle (just before the line-pre-redraw hooks are called). This would probably be the right time for z-sy-h as well.

Yes.

> So setting that flag would be kind of like saying "I want this highlight to be around for the next redraw cycle, but remove it after that".
> 
> I'm a little apprehensive to give up control of when exactly the entries are removed, but I'm struggling to come up with any real justification for that apprehension. Would there be cases where someone would want a highlight to stick around until they manually remove it?

Who's "someone"?  A user?  A plugin author?  Whoever it is, couldn't they add their entry without `autoremove=yes` (or remove the `autoremove=yes` annotation from an existing entry)?

If you're getting at making `autoremove=yes` the default, wouldn't that be backwards incompatible?

> > Okay, so you're proposing to encapsulate adding a single entry and removing a single entry. However, operations such as "remove all my entries", "remove all entries", "enumerate entries", etc., would not be encapsulated, so they'd only be possible to implement by directly accessing the array.
> 
> That's basically what I'm thinking. You could still support "remove all my entries" by passing multiple tokens to `zle -h -d`. And if there's some way to return multiple entries in REPLY, we could add multiple entries using `zle -h` as well.

There is: by convention, `$REPLY` is a scalar and `$reply` is an array.

> "remove all entries" could be supported by allowing a pattern arg to `zle -h -d` and passing something like `zle -h -d *`. "enumerate entries" could be `zle -h -l` or something. So I think we could wrap the whole thing up in the `zle` interface, but it will be a bit more clunky for those use cases where the param makes more sense.
> 

It's certainly possible to encapsulate these other operations as well.

> > Doesn't that somewhat negate the benefits of encapsulation, at least at the zle level?
> 
> Somewhat yeah. The benefit we'd get from adding `zle -h` is being able to have side effects when a highlight is added (setting REPLY to a unique identifier for that highlight).

Do you see any other use-cases besides setting $REPLY as mentioned?

> I'm thinking we keep `region_highlight` around in its current form mainly for backward compatibility, but there are some other options:
> 
> 1. Deprecate `region_highlight` and encapsulate highlight behavior entirely in new `zle -h`.
> 2. Make `region_highlight` read-only and encapsulate adding/removing highlights in `zle -h`. (Isn't this kind of like existing params like `$widgets` that are modified by builtins but available as read-only params for convenience?)
> 

So, the code you're proposing is:

```zsh
local s="P0 20 bold"
zle -h -- $s
typeset -g _z_asug_cookie=$REPLY
```

and the code I'm proposing is:

```zsh
local s="P0 20 bold"
typeset -g _z_asug_cookie=:z-asug:$RANDOM:
region_highlight+=( "$s owner=$_z_asug_cookie" )
```

And the main difference is that your proposal involves designing, implementing, and documenting a new API.  If the old API remains fully supported, the first proposal would increase zle's maintenance burden; and if we don't, the existing UI would have to be deprecated [which has a cost in man-hours too], and subsequently plugin maintainers would have to update their code.

I am sorry, but I do not consider separating the generation-of-nonces concern from z-asug to be sufficient justification for imposing these costs on zle and other plugins.

What's so foreboding about the second alternative?  All you said was:

> > > I don’t care what the identifier value is, so having to specify some value for it is just extra work for me to do.

I see that that the second alternative's design is not in the [first normal form](https://en.wikipedia.org/wiki/First_normal_form), and that it allows stuff like random write access (`region_highlight[$i]=$foo`), both of which we might have done differently if we'd been designing this from scratch — but we aren't designing this from scratch.  We're designing a change to an existing API with an existing ecosystem of API consumers around it.  In such cases, it's good practice to opt for small changes in preference to massive rewrites, where possible.

If it helps, we could add some sort of magic parameter that expands to a different value every time it's referenced: perhaps $AUTOINCREMENT (expands to 0, then 1, then 2, then 3, …) or $NONCE (value only guaranteed to be unique; maybe $UUIDGEN after `uuidgen(1)`).

Or you could propose that the `zle-h` (sic) function I posted a few comments ago be cleaned up, extended, and added to the distribution as a standard autoloadable function.  Something like that could be useful, because it could solve the "not in the first normal form" problem centrally, for both reads and writes. It could also take care of inventing nonces.

> > An encapsulation, in this case, would seem to be another API entry point to maintain, but without the benefit of being able to change the implementation without changing the API callers?
> 
> I think the implementation could technically change as long as the API contracts are kept intact. The API contracts being the details of both `zle -h` builtin and `region_highlight` param. For example, the algorithm to generate the unique token for each new  `zle -h` highlight could be altered without any changes required to clients.

Well, yes, that _is_ an example of encapsulating an implementation detail.  What I fail to see is why that detail needs to be encapsulated in the first place.  Do you expect best practices in the area of nonce-generation to change in the coming years?

Next steps:

- Sound the mailing list about `owner=$owner` v. `autoremove=yes`.

- @ericfreese What's so foreboding about that second code snippet above?

Cheers!

okapia:
> (Aside: I'll change "owner" in the PR to some more generic term before posting/merging this.)

"id" perhaps. Is nice and short. Could document a convention of prefixing it with a string to identify the owning plugin.

okapia:
> Would there be cases where someone would want a highlight to stick around until they manually remove it?

Most definitely yes. Most of my uses of region_highlight fall into this category. The vi-exchange plugin of mine that is public needs the highlight to stay around.

An autoremove, entry seems reasonable. (the shorter name might be nice). Could be good to have the value make it clear what the remove criteria would be - also useful for future extensions, e.g. `autoremove=key` for when another key is pressed. Or even `autoremove=0.5` for a ½ second.

okapia:
I can't say I really like the `zle -h` suggestion. I like the visibility and transparency that comes with having something I already understand and know how to manipulate like a variable. And I don't see how it really makes this any easier.

If it generates an id for you in `$REPLY`, a plugin that lives from one widget invocation to the next needs to store that id in a global variable. And as global variable names clash, what have you gained over hardcoding your own ids. And they'll be less meaningful, I'd rather see `id=vi-exchange:selected` than `id=23fe1ca`.

danielshahaf:
> An autoremove, entry seems reasonable. (the shorter name might be nice).

What "shorter name" are you referring to? None has been proposed.

> Could be good to have the value make it clear what the remove criteria would be - also useful for future extensions, e.g. autoremove=key for when another key is pressed. Or even autoremove=0.5 for a ½ second.

+1

okapia:
> What "shorter name" are you referring to? None has been proposed.

That was a typo or similar. I meant _a_ not _the_.

Ideas that come to mind include, "expire", "expiry", "lifetime", "remove", "scope", "transient", "tmp"

okapia:
> I was thinking of `rhid` (initialism of "`region_highlight` id" ) for greppability reasons.

I'm not especially fond of such identifying initialisms. It's only used within `region_highlight`. For grep, `region_highlight.*id` ought to work often enough (barring word wrapping) and for a particular plugin, `id=z-s-h` may work.

danielshahaf:
> Ideas that come to mind include, "expire", "expiry", "lifetime", "remove", "scope", "transient", "tmp"

`expiry` sounds right.  It's a somewhat more abstract than `autoremove`.

> I'm not especially fond of such identifying initialisms. It's only used within `region_highlight`. For grep, `region_highlight.*id` ought to work often enough (barring word wrapping) and for a particular plugin, `id=z-s-h` may work.

In my experience, self-descriptive names are useful.  See, for example, https://subversion.apache.org/issue3744.  That's why I'm not that fond of bare `id`.

But I get that `rhid` is a bit cryptic.  Kinda comes with the territory of being short, though.  Anyway, how about, say, `stroke` or `stroke_id`?

Do we need the `owner=` functionality at all if we have the `autoremove` (or `expiry`) functionality?

---

Are there any outstanding concerns with the functionality, other than the names of the API entry points?

okapia:
> In my experience, self-descriptive names are useful. See, for example, https://subversion.apache.org/issue3744. That's why I'm not that fond of bare `id`.

With the scope limited to the region_highlight array, I can't see there being any confusion. 

> But I get that `rhid` is a bit cryptic. Kinda comes with the territory of being short, though. Anyway, how about, say, `stroke` or `stroke_id`?

"stroke" is too suggestive of a movement in it's meaning which I don't think applies here. It may have subtle US/UK differences in meaning. How about "mark". A thesaurus also turns up "tag", "stamp", "label" and "badge". In a somewhat different vein, "memo" or "key".

> Do we need the `owner=` functionality at all if we have the `autoremove` (or `expiry`) functionality?

Checking back over the discussion, I don't think you exactly fleshed out when auto-removal would occur. It perhaps removes the need from z-s-h and any widget that re-applies things any time the buffer changes. But for longer persisting things like vi-exchange, it'd still be useful.

danielshahaf:
> > In my experience, self-descriptive names are useful. See, for example, https://subversion.apache.org/issue3744. That's why I'm not that fond of bare `id`.  
> 
> With the scope limited to the region_highlight array, I can't see there being any confusion. 
> 

Yeah, but a unique name would be more greppable.  `grep -R rhid` is in practice going to have near-100% specificity and sensitivity; that's not the case for either `grep -Rw id` or `grep -R 'region_highlight.*id'`.  [This goes not only for `rhid` but also for any other sufficiently-unique term.]

> > But I get that `rhid` is a bit cryptic. Kinda comes with the territory of being short, though. Anyway, how about, say, `stroke` or `stroke_id`?  
> 
> "stroke" is too suggestive of a movement in it's meaning which I don't think applies here. It may have subtle US/UK differences in meaning. How about "mark". A thesaurus also turns up "tag", "stamp", "label" and "badge". In a somewhat different vein, "memo" or "key".
> 

I was thinking of the noun sense of 'stroke', but yeah, something else.  'label' and 'memo' both sound good.

> > Do we need the `owner=` functionality at all if we have the `autoremove` (or `expiry`) functionality?  
> 
> Checking back over the discussion, I don't think you exactly fleshed out when auto-removal would occur.

Autoremoval should take place at the start of the following redraw cycle (the cycle _after_ the one zle-line-pre-redraw is running as part of), I suppose?

> It perhaps removes the need from z-s-h and any widget that re-applies things any time the buffer changes. But for longer persisting things like vi-exchange, it'd still be useful.

Ack usefulness.  So, I think the subtotal up to this point is to add both memo= (née owner=) and expiry=next-redraw (née autoremove=yes)?  If so, let's take this proposal to -workers@.

ericfreese:
> If you're getting at making `autoremove=yes` the default, wouldn't that be backwards incompatible?

No, was just wondering if we would still need the "owner" solution. Sounds like we do.

> Do you see any other use-cases besides setting $REPLY as mentioned?

No that's the only one I'm seeing for now.

> I see that that the second alternative's design is not in the [first normal form](https://en.wikipedia.org/wiki/First_normal_form), and that it allows stuff like random write access (`region_highlight[$i]=$foo`), both of which we might have done differently if we'd been designing this from scratch — but we aren't designing this from scratch.  We're designing a change to an existing API with an existing ecosystem of API consumers around it.  In such cases, it's good practice to opt for small changes in preference to massive rewrites, where possible.

Makes sense. I figured this would be the outcome but thought I'd give my 2 cents anyway.

> If it helps, we could add some sort of magic parameter that expands to a different value every time it's referenced: perhaps $AUTOINCREMENT (expands to 0, then 1, then 2, then 3, …) or $NONCE (value only guaranteed to be unique; maybe $UUIDGEN after `uuidgen(1)`).

This is an interesting idea. Probably not necessary for now, but I'll keep it in mind for the future.

> What's so foreboding about that second code snippet above?

Nothing in particular. I was just thinking it might be a nice feature to implement so that clients (plugins, users, etc.) wouldn't need to worry about implementing the nonce-generation themselves.

> I think the subtotal up to this point is to add both memo= (née owner=) and expiry=next-redraw (née autoremove=yes)?

Do we need both? I would be fine with just memo/owner for now.

danielshahaf:
> > I think the subtotal up to this point is to add both memo= (née owner=) and expiry=next-redraw (née autoremove=yes)?  
> 
> Do we need both? I would be fine with just memo/owner for now.

Good point.  Yes, let's add just memo=, then.  Unless implementing expiration in zle itself would be a noticeable performance gain?  (@psprint?)

danielshahaf:
Thread: [workers/46001](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=46001)

Src/Zle/zle_refresh.c
danielshahaf: Should be resolved by the just-pushed commits.  (The hex commit id's will be changed later when I squash them, so I'm not pointing to a specific commit.)

danielshahaf: New tests added for this.

danielshahaf:
Push some commits that rename `owner=` to `memo=` and implement forwards compatibility.  I'll probably squash the lot before I post to -workers@, but I'm posting them here unsquashed for ease of review for those who have reviewed the previous (first) revision.

danielshahaf:
Two TODO comments remain:

```diff
+           /* ### TODO: Consider optimizing the common case that memo_start to
+            *           end-of-string is entirely ASCII */
+           while (1) {
```

```diff
+           /* TODO: save rhp->memo here? */
            if (zlemetaline) {
                newrhp->start = rhp->start_meta;
```

Src/Zle/zle_refresh.c
```diff
@@ -502,7 +512,44 @@ set_region_highlight(UNUSED(Param pm), char **aval)
...
+		int nbytes;
+		convchar_t c = unmeta_one(i, &nbytes);
+
+		if (c == '\0' || c == ',' || inblank(c)) {
+		    memo_end = i;
+		    break;
+		} else
+		    i += nbytes;
+	    }
+	    rhp->memo = ztrduppfx(memo_start, memo_end - memo_start);
```
danielshahaf, line 550: Where's the corresponding free()?

danielshahaf: Pushed.

danielshahaf:
Posted this PR as of 23d1edce5d607addda4e38dc124485a0a599ce96, squashed, to -workers@ in [workers/46068](http://www.zsh.org/cgi-bin/mla/redirect?WORKERNUMBER=46068).

danielshahaf:
Merged in dd6e702ee49c7292c39037843b1b1b2b080f9fda.  Thanks for all the reviews and testing!

I'll go back to https://github.com/zsh-users/zsh-syntax-highlighting/issues/418 and see what can be done to move it forward, now.

PR #58, opened 2020-07-03
Add instructions for the platform dependency

jnooree:
This add notes for the platform dependency of the module, related to the non-standard regular expressions.

### DIFF
--- a/Doc/Zsh/mod_regex.yo
+++ b/Doc/Zsh/mod_regex.yo
@@ -28,13 +28,15 @@ If tt(BASH_REMATCH) is set, then the array tt(BASH_REMATCH) will be set
 instead of tt(MATCH) and tt(match).
 
 Note that the tt(zsh/regex) module logic relies on the host system. The 
-same var(expr) and var(regex) pair could result in different evaluations,
-especially if var(regex) with non-standard syntax is given, thus should
-be avoided.
+same var(expr) and var(regex) pair could produce different results on different
+platforms if a var(regex) with non-standard syntax is given.
 
-For example, the escape character tt(\b) representing a word boundary is
-undefined in the POSIX extended regular expression standard. It works as
-expected in GNU-Linux, but does not work (translated as a character tt(b))
-in BSD-based platforms.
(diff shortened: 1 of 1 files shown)


danielshahaf:
Thanks!  WDYT of these changes:

```yodl
Note that the tt(zsh/regex) module logic relies on the host system. The 
same var(expr) and var(regex) pair could produce different results on different
platforms if a var(regex) with non-standard syntax is given.

For example, no syntax for matching a word boundary is defined in the POSIX
extended regular expression standard. GNU libc and BSD libc both provide such
syntaxes as extensions (tt(\b) and tt([[:<:]])/tt([[:>:]]) respectively), but
neither of these syntaxes is supported by both of these implementations.

Refer to the manref(regcomp)(3) and manref(re_format)(7) manual pages on your
system for locally-supported syntax.
```

?

<details>
<summary>
wdiff to your version
</summary>

```wdiff
Note that the tt(zsh/regex) module logic relies on the host system. The 
same var(expr) and var(regex) pair could [-result in-] {+produce different results on+} different [-evaluations,
especially-]
{+platforms+} if {+a+} var(regex) with non-standard syntax is [-given, thus should
be avoided.-] {+given.+}

For example, [-the escape character tt(\b) representing-] {+no syntax for matching+} a word boundary is
[-undefined-] {+defined+} in the POSIX
extended regular expression standard. [-It works-] {+GNU libc and BSD libc both provide such
syntaxes+} as
[-expected in GNU-Linux,-] {+extensions (tt(\b) and tt([[:<:]])/tt([[:>:]]) respectively),+} but [-does not work (translated as a character tt(b))
in BSD-based platforms.-]
{+neither of these syntaxes is supported by both of these implementations.

Refer to the manref(regcomp)(3) and manref(re_format)(7) manual pages on your
system for locally-supported syntax.+}
```
</details>

jnooree:
Thanks for improvements. I'll add those!

danielshahaf:
Looks good to me!

danielshahaf:
Squashed and merged in 89618fe37d0c.  Thanks for the patch!

PR #60, opened 2020-07-08
Completion: Fix completion for _zsh_dataset containing spaces

mafredri:
I don't know if this is the case for all ZFS implementations, but spaces are allowed on OpenZFS / Linux in dataset names. In the current implementation, the trailing part of these datasets will appear as a new completion entry and the real path can never be completed. Since we're already passing the `-H` flag which uses tab-separation and each dataset appears on it's own line, we just need to limit word splitting to newlines.

Reproduction:

```
# zfs create rpool/storage/my\ data
# zfs list [tab]
[ filesystem/volume/snapshot/path ]
data rpool
```


okapia:
Looks good to me. I've pushed this in aa60dd3. Thanks for contributing the fix.

PR #61, opened 2020-07-14
_parameters: fix non-local variable

oxiedi:
Something to reproduce:

    % zstyle -e ':completion:*' matcher-list 'local -i i=0; reply=( "" )'
    % zstyle ':completion:*' fake-parameters 'ENVVAR:scalar'
    % ENV<Tab>
    _parameters:22: bad math expression: ':' without '?'


danielshahaf:
Thanks.  Merged in 2516da75a91f1daead579c5c5004d89ff9e33a3f.

PR #62, opened 2020-07-15
_date_formats: return proper exit status

oxiedi:
It fixes messed up completion for the following:

    % zstyle ':completion:*' matcher-list '' ''
    % date +<Tab>


okapia:
I think the status from _describe is what needs to be returned. _message doesn't return anything especially useful. So I've tweaked the patch before committing (in 2b2353fc).

Feel free to reopen and let me know if you think this isn't right. Thanks again for the patch and apologies for our tardiness in addressing it.

PR #63, opened 2020-07-20
_values: fix regression caused by cd6fd2b0a3641774e7854ff8298d1d82643c4b4a

oxiedi:


danielshahaf:
@oxiedi Hi, thanks for these two PRs.  Just wanted to say we might be a bit slower than usual in responding to these due to summer holidays and behind-the-scenes workloads.

okapia:
Sorry for the delay. I finally had time to verify the patch and work out how to reproduce the issue. It looks right and I have merged the fix (in df7fd132). However, I'm not convinced by the test case including the zstyle removal step and the order it does things in so I've left that out. I thing a test that sets the style would make more sense. I don't think it is especially important but feel free to followup with just the test case if you think it needs more consideration.

Thanks for the patch.

PR #65, opened 2020-09-21
Update _nmap

mengbo:
Nmap commands that previously included an underscore (--max_rtt_timeout, --send_eth, --host_timeout, etc.) have been renamed to use a hyphen in the preferred format.


danielshahaf:
+1. Will merge when at my workstation.

danielshahaf:
Merged in 59a451abae44 with a tweaked log message.  Thanks!

PR #66, opened 2020-11-29
Add more commands to _todo.sh

DCsunset:
I have added more commands to `_todo.sh`.


okapia:
Thanks for the contribution. Sorry for not having looked over it sooner.

Merged in f59b6f6d.

PR #67, opened 2020-11-30
Dot Etc Path Adjustments

benmelz:


danielshahaf:
For future reference, see https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html.

benmelz:
This was meant to be a PR on a fork of mine, sorry about that.

PR #68, opened 2020-12-10
_man: support the man-db syntax <name>.<section> 

taiyu-len:
``@todo We don't support the man-db syntax <name>.<section> (e.g., `ls.1`)``

This adds support for the above todo.  
it is enabled using append-sections. `zstyle ':completion:*:manuals.*' append-sections true`  
insert-sections takes priority so it must be disabled/not enabled in the first place in order to use.

whoops, force pushed some fixes because i made a dumb mistake due to not testing the committed code,  
and switched to `-s` for `compadd` since that adds a space after completion unlike `-S`.

### DIFF
--- a/Completion/Unix/Command/_man
+++ b/Completion/Unix/Command/_man
@@ -6,7 +6,6 @@
 # - We assume that Linux distributions are using either man-db or mandoc
 # - @todo Would be nice to support completing the initial operand as a section
 #   name (on non-Solaris systems)
-# - @todo We don't support the man-db syntax <name>.<section> (e.g., `ls.1`)
 # - @todo We don't support the man-db feature of 'sub-pages' — that is, treating
 #   pairs of operands like `git diff` as `git-diff`
 # - @todo Option exclusivity isn't super accurate
@@ -415,7 +414,7 @@ _man() {
 }
 
 _man_pages() {
-  local pages sopt
+  local pages sopt tmp
 
   # What files corresponding to manual pages can end in.
   local suf='.((?|<->*|ntcl)(|.gz|.bz2|.z|.Z|.lzma))'
@@ -444,13 +443,20 @@ _man_pages() {
   # `POSIX.1.5'.
 
   [[ $variant = solaris* ]] && sopt='-s '
-  if ((CURRENT > 1 || noinsert)) ||
-      ! zstyle -t ":completion:${curcontext}:manuals.$sect_dirname" insert-sections
-  then
-    compadd "$@" - ${pages%$~suf}
-  else
-    compadd "$@" -P "$sopt$sect_dirname " - ${pages%$~suf}
+  if ! ((CURRENT > 1 || noinsert)); then
+    zstyle -s ":completion:${curcontext}:manuals.$sect_dirname" insert-sections tmp
   fi
+  case "$tmp" in
+    prepend|true|on|yes|1)
+      compadd "$@" -P "$sopt$sect_dirname " - ${pages%$~suf}
+    ;;
+    suffix)
+      compadd "$@" -s ".$sect_dirname" - ${pages%$~suf}
+    ;;
+    *)
+      compadd "$@" - ${pages%$~suf}
+    ;;
+  esac
 }
 
 _man "$@"
--- a/Completion/Zsh/Command/_zstyle
+++ b/Completion/Zsh/Command/_zstyle
@@ -62,6 +62,7 @@ styles=(
   ignore-parents         c:ignorepar
   ignored-patterns	 c:
   insert-ids             c:insert-ids
+  insert-sections	 c:insert-sections
   insert-tab             c:bool
   insert-unambiguous	 c:insunambig
   keep-prefix		 c:keep-prefix
@@ -522,6 +523,11 @@ while (( $#state )); do
     (insert-ids)
       _wanted values expl 'when to insert process IDs' \
           compadd - menu single longer
+      ;;
+
+    (insert-sections)
+      _wanted values expl 'where to insert man page section' \
+          compadd - true false prepend suffix
       ;;
 
     (fake-files)
--- a/Doc/Zsh/compsys.yo
+++ b/Doc/Zsh/compsys.yo
@@ -1838,13 +1838,17 @@ user is longer than the common prefix to the corresponding IDs.
 kindex(insert-sections, completion style)
 item(tt(insert-sections))(
 This style is used with tags of the form `tt(manuals.)var(X)' when
-completing names of manual pages. If set to `true' and the var(X) in the
-tag name matches the section number of the page being completed, the
-section number is inserted along with the page name. For example, given
+completing names of manual pages. If set and the var(X) in the tag name matches
+the section number of the page being completed, the section number is inserted
+along with the page name. For example, given
 
 example(zstyle ':completion:*:manuals.*' insert-sections true)
 
 tt(man ssh_<TAB>) may be completed to tt(man 5 ssh_config).
+
+The value may also be set to one of `tt(prepend)', or `tt(suffix)'.
+`tt(prepend)' behaves the same as `true' as in the above example, while
+`tt(suffix)' would complete tt(man ssh_<TAB>) as tt(man ssh_config.5).
 
 This is especially useful in conjunction with tt(separate-sections), as
 it ensures that the page requested of tt(man) corresponds to the one


danielshahaf:
@taiyu-len Thanks for the patch. Although I haven't time to review it right now, I'd like to point out that the `_zstyle` completion would want to be updated before this is merged. Feel free to wait for further feedback before making this change.

okdana:
I tested this briefly and it seems OK functionally on macOS (and the logic makes sense afaict).

I'm not sure about the configuration though. It's a bit weird to have two styles with very similar names that control the same kind of behaviour, but are conflicting, and have a precedence that's basically arbitrary. Would it be better to make `insert-sections` take string values like `prepend` and `append` (where `true` is the same as `prepend` for BC)? Maybe Daniel or Oliver has an opinion on the UX.

If we *do* want to do that, i think you would change the check for the original behaviour to be a pattern match like `(prepend|1|on|true|yes)`, since it will no longer do the boolean conversion for you. And obv you'd have to update the documentation — look around that file and you'll see several other examples where the style takes both boolean and string values.

If we *don't* want to do that, i would suggest at least changing the new documentation section so it isn't so copy/paste from the other one. You can have it say 'See `insert-sections` for more information' or w/e rather than duplicating all of that text.

taiyu-len:
Thanks, using a single style name is certainly cleaner all around. Im not too familiar, so wasnt sure how best to do it.  
the relevant logic could then be
```bash
  if ! ((CURRENT > 1 || noinsert)); then
    zstyle -s ":completion:${curcontext}:manuals.$sect_dirname" insert-sections tmp
  fi
  case "$tmp" in
    prepend|true|on|yes|1)
      compadd "$@" -P "$sopt$sect_dirname " - ${pages%$~suf}
    ;;
    append)
      compadd "$@" -s ".$sect_dirname" - ${pages%$~suf}
    ;;
    *)
      compadd "$@" - ${pages%$~suf}
    ;;
  esac
```


I can clean up the rest and push changes whenever

okdana:
ty, that seems good to me at first glance, but we can see if anyone else has an opinion

danielshahaf:
> I'm not sure about the configuration though. It's a bit weird to have two styles with very similar names that control the same kind of behaviour, but are conflicting, and have a precedence that's basically arbitrary. Would it be better to make `insert-sections` take string values like prepend and append (where `true` is the same as `prepend` for BC)?

+1, with the nitpick that `prepend`/`append` terminology implies symmetry which isn't there, unlike, say, `prepend-word`/`append-after-a-dot-separator`.

okapia:
> I'm not sure about the configuration though. It's a bit weird to have two styles with very similar names that control the same kind of behaviour, but are conflicting, and have a precedence that's basically arbitrary. Would it be better to make `insert-sections` take string values like `prepend` and `append` (where `true` is the same as `prepend` for BC)? Maybe Daniel or Oliver has an opinion on the UX.

I'm not sure I'm especially keen on overloading `separate-sections` rather than have it do one thing only. Note also that the style is also used in `_dict_words`. I set it with a context that isn't specific to man. Also, doesn't this have far more in common with the existing [undocumented] `insert-sections` style. Let that take a `suffix` value for appending. Support for .<section> is by no means universal across man implementations. I wasn't especially aware of it. Seems to be groff specific. Could be useful if we can detect that for users so they don't need different styles for BSD/Solaris/void linux vs. Redhat/macOS.

Before removing the `TODO` item, consideration should be given to users starting a section. I would quite like sections *not* to be included by default but I should be able to complete sections after, e.g. `man open.`
dots in man pages are rare but do occur. Completing for rpc(3) and rpc.lockd(8) should not be adversely affected. Checking for dots, I'm suspicious that some (.lua, .tcl86) are meant to be sections. As part of this, a dot should be added to the auto-suffix removal list (`-r` option) when adding man pages without a section.

With `-S`, you should include the space to keep it in the suffix. In this case, I would argue that `-s` is correct anyway but should be combined with `-M  'r:|.=* r:|=*'` to match characters preceding the dot and at the end of the section.

okdana:
> I'm not sure I'm especially keen on overloading `separate-sections` rather than have it do one thing only. Note also that the style is also used in `_dict_words`. I set it with a context that isn't specific to man. Also, doesn't this have far more in common with the existing [undocumented] `insert-sections` style. Let that take a `suffix` value for appending.

That's what's being suggested, yes. `separate-sections` would not be affected at all.

> Support for `.<section>` is by no means universal across man implementations. I wasn't especially aware of it. Seems to be groff specific. Could be useful if we can detect that for users so they don't need different styles for BSD/Solaris/void linux vs. Redhat/macOS.

Right, it's only supported by (recent versions of?) man-db afaik — and it's an alternative syntax to the usual `man 1 ls` style, it's not required to specify the section. I'm unsure if there's a specific reason to use it over the more traditional style. I can only assume that some Linux users are accustomed to the aesthetics of it...?

I guess i should have asked before: @taiyu-len, is that why you're suggesting the change? Or do you know of a reason you would *require* this syntax over the other one?

> Before removing the `TODO` item, consideration should be given to users starting a section. I would quite like sections _not_ to be included by default but I should be able to complete sections after, e.g. `man open.`
> dots in man pages are rare but do occur. Completing for rpc(3) and rpc.lockd(8) should not be adversely affected. Checking for dots, I'm suspicious that some (.lua, .tcl86) are meant to be sections. As part of this, a dot should be added to the auto-suffix removal list (`-r` option) when adding man pages without a section.

I wasn't even thinking about that. Well, i wasn't when i saw this PR — i'm pretty sure it's exactly what i meant when i wrote the todo, since as i mentioned this syntax is not necessary to make `insert-sections` work well with man-db.

For actually completing sections after a dot on the command line, like you said, we shouldn't need any configuration stuff; we would just determine its applicability the same way we do with the option specs. And that is probably a more useful change.

danielshahaf:
> I'm unsure if there's a specific reason to use it over the more traditional style. 

In a word: statelessness. IIRC, `man 3 foo printf` would open `printf(3)`, whereas `man printf` would open `printf(1)`. Thus, the meaning of the last positional argument depends on the ones before it: the meaning of `man -- "$@" printf` depends on `"$@"`. This is a form of action at a distance, which is an antipattern and addressed by a (normalized) syntax that avoids state-keeping.

okapia:
Are there any outstanding questions for this PR? If I understand correctly it is waiting on a "clean up the rest and push changes to the pr" step. Would be good to get this finished and merged.

taiyu-len:
My bad, was waiting to see if more would be discussed then ended forgotting to get around to it.

the only difference from the above code I posted is that it uses suffix rather then append for the setting as suggested above.

anyway, pushed it now. there are two things im not sure how to handle.

first, mentioned in the post above `but should be combined with -M 'r:|.=* r:|=*' to match characters preceding the dot and at the end of the section.`
i tried working that into it, however since the words passed into compadd have the `.N.*` suffix removed it doesnt work for sections. and modifying the suf variable used in `${pages%$~suf}` causes it to display the .N in each completion option which I assume is undesirable.

second is when completing words such as `printf.` and `printf.3`.
the ideal behavior in the first case is to show the completion for `printf`, however it treats the suffix as an error and gives corrections instead.
likewise, for the second case, it treats it as an error and shows corrections, when ideally it should just insert a space since its already completed.

okapia:
Completion doesn't typically expect to find the cursor within the suffix part of a completion match. With `-s`, matching does include the suffix but the cursor should be before it. One approach would be to do something like:

```
      disp=( ${pages%$~suf} )
      compadd "$@" -M 'r:|.=* r:|=*' -d disp - ${^disp}.$sect_dirname
```

That also needs `local -a disp` at the top of the function. But this approach uses more memory because a separate display string is stored per match. The alternative is to check for a dot in the prefix and use a different `compadd` invocation when it is found.

I'd also be inclined to apply the new behaviour unconditionally where a dot is already there on the command-line. For example, the following could be used:

```
    if [[ -prefix 1 *. && $variant = linux* ]]; then
      tmp=suffix
    else
```

Where the zstyle lookup goes in the else condition. I'm not sure if linux in variant is a complete way to detect support for this. It could also be good to allow different values of the style to say you want suffix on a platform like Linux that supports it but otherwise prepend.

Right at the beginning, Daniel suggested updating `_zstyle` for the new style values. Commit ceb25d94 that I made yesterday is very similar and could provide a template for what to do.

Let me know, I'm happy to merge the latest change as-is or if you want, push a further update with some of the ideas just outlined. Either way, let me know - would be good to get this finalised.

Thanks for reacting promptly to my ping from yesterday.

taiyu-len:
Thanks, I added the zstyle compltion bit for insert-sections, as suggested.
Regarding the unconditionally using suffix when a period is present, I decided not to add that as it changes behavior for manpage names of anything containing a period.
And as for the display string, I think its better to leave the decision on whether adding that extra work is worth it to someone who knows better then I.

otherwise seems good to merge as i see it.

okapia:
I've merged this in 3714ee0b.

Thanks for your contribution.

PR #69, opened 2020-12-15
Restore old value of vi repeat

xPMo:
Should fix romkatv/powerlevel10k#1042 and https://zsh.org/workers/47432.

### DIFF
--- a/Src/Zle/zle_main.c
+++ b/Src/Zle/zle_main.c
@@ -1065,6 +1065,7 @@ redrawhook(void)
 	int saverrflag = errflag, savretflag = retflag;
 	int lastcmd_prev = lastcmd;
 	int old_incompfunc = incompfunc;
+	int old_viinrepeat = viinrepeat;
 	char *args[2];
 	Thingy lbindk_save = lbindk, bindk_save = bindk;
 
@@ -1079,6 +1080,7 @@ redrawhook(void)
 	incompfunc = 0;
 	execzlefunc(initthingy, args, 1, 0);
 	incompfunc = old_incompfunc;
+	viinrepeat = old_viinrepeat;
 
 	/* Restore errflag and retflag as zlecallhook() does */
 	errflag = saverrflag | (errflag & ERRFLAG_INT);
--- a/Src/Zle/zle_main.c
+++ b/Src/Zle/zle_main.c
@@ -1062,10 +1062,11 @@ redrawhook(void)
     Thingy initthingy;
     if ((initthingy = rthingy_nocreate("zle-line-pre-redraw"))) {
 	/* Duplicating most of zlecallhook() to save additional state */
-	int saverrflag = errflag, savretflag = retflag;
-	int lastcmd_prev = lastcmd;
-	int old_incompfunc = incompfunc;
-	int old_viinrepeat = viinrepeat;
+	int errflag_save = errflag;
+	int retflag_save = retflag;
+	int lastcmd_save = lastcmd;
+	int incompfunc_save = incompfunc;
+	int viinrepeat_save = viinrepeat;
 	char *args[2];
 	Thingy lbindk_save = lbindk, bindk_save = bindk;
 
@@ -1079,12 +1080,12 @@ redrawhook(void)
 	 */
 	incompfunc = 0;
 	execzlefunc(initthingy, args, 1, 0);
-	incompfunc = old_incompfunc;
-	viinrepeat = old_viinrepeat;
+	incompfunc = incompfunc_save;
+	viinrepeat = viinrepeat_save;
 
 	/* Restore errflag and retflag as zlecallhook() does */
-	errflag = saverrflag | (errflag & ERRFLAG_INT);
-	retflag = savretflag;
+	errflag = errflag_save | (errflag & ERRFLAG_INT);
+	retflag = retflag_save;
 
 	unrefthingy(initthingy);
 	unrefthingy(lbindk);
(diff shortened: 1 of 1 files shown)


xPMo:
@romkatv Could you bump this again on zsh-workers? I'm unsure on ML etiquette (plus I need to get aerc setup so I can `git send-email` correctly).

xPMo:
Also, I'm adding another commit which renames things in that function to be more consistent: (sav[ident], old_[ident], [ident]_prev, [ident]_save -> [ident]_save). It was just bugging me.

romkatv:
> @romkatv Could you bump this again on zsh-workers? I'm unsure on ML etiquette

It's reasonable to bump the thread. Simply Reply-All, mention that you are the author of the patch and ask Peter whether he (or perhaps someone else) can merge it. Technically I can merge it but I don't know that part of the code well enough.

> Also, I'm adding another commit which renames things in that function to be more consistent: (sav[ident], old_[ident], [ident]_prev, [ident]_save -> [ident]_save). It was just bugging me.

I wouldn't send this patch.

danielshahaf:
> > Also, I'm adding another commit which renames things in that function to be more consistent: (sav[ident], old_[ident], [ident]_prev, [ident]_save -> [ident]_save). It was just bugging me.
> 
> I wouldn't send this patch.

Please state not only your conclusion but also its rationale.

xPMo:
> It's reasonable to bump the thread. Simply Reply-All, mention that you are the author of the patch and ask Peter whether he (or perhaps someone else) can merge it. Technically I can merge it but I don't know that part of the code well enough.

Okay, I found the message in my archived, I'll do that.

> I wouldn't send this patch.

Understandable, it doesn't fix an issue and clogs up git blame.

romkatv:
> > I wouldn't send this patch.

> Please state not only your conclusion but also its rationale.

Your guess is as good as mine.

xPMo:
Merged in `5d14b6eb7`

romkatv:
Thanks for following through.

danielshahaf:
> > Please state not only your conclusion but also its rationale.
> 
> Your guess is as good as mine.

I don't see what you would be guessing here.  In any case, Roman, my point, Roman, is that the target audience — PR authors, fellow developers, and lurkers — shouldn't have to guess the reviewer's reasons.

Also, spelling out a reason can unearth new conclusions.  (Compare https://www.paulgraham.com/essay.html.)

The presumed reasons have been spelled out now by @xPMo, so I don't think there's much else to say.

romkatv:
> In any case, Roman, my point, Roman, is that the target audience — PR authors, fellow developers, and lurkers — shouldn't have to guess the reviewer's reasons.

I don't consider myself a reviewer of this patch. I just forwarded it to the mailing list.

If I were in xPMo's shoes -- that is, having sent a small patch that fixes a bug that affects me and not receiving any feedback on it -- I wouldn't send an additional patch that renames variables. If I imagine that I arrived at this course of action by reason, it might look like this.

- The original patch is small and uncontentious. It fixes the bug without doing anything else.
- The additional patch has pros (cleaner code) and cons (code churn, history noise). I don't know how much zsh maintainers value clean code vs fewer changes and simpler history.
- The original patch didn't get any feedback because people are busy. It doesn't take much time to review and merge the patch but the value of the patch is probably not great for those who need to do the work.
- If I send the second patch, it'll demand more work from people who are already busy.
- If that patch becomes contentious, it'll reduce the chance that my first patch (the one I actually care about) will be merged because the two might be seen as a whole package.
- Even if the second patch is not seen as contentious, it still won't increase the chance the first patch is merged, so it doesn't help my main goal of having the bug fixed.

I didn't actually think like this when I wrote that I wouldn't send the second patch. I'm just guessing at how my mind came up with that decision. I thought that sharing what I would do could be useful to xPMo. It hasn't occurred to me that I may be seen as a reviewer and/or maintainer of zsh. I though I'm commenting on https://github.com/romkatv/powerlevel10k/issues/1042. Only now I've realized that this thread is on zsh-users/zsh. I wouldn't write such a short and personal remark as a reviewer/maintainer.

> Also, spelling out a reason can unearth new conclusions.

True. It's work though.

danielshahaf:
> > In any case, Roman, my point, Roman, is that the target audience — PR authors, fellow developers, and lurkers — shouldn't have to guess the reviewer's reasons.
> 
> I don't consider myself a reviewer of this patch. I just forwarded it 
> to the mailing list.

Ah, I see.  Your recommendation not to send the second patch _was_ a review, though, notwithstanding that your pointing to the list wasn't.

> If I were in xPMo's shoes -- that is, having sent a small patch that 
> fixes a bug that affects me and not receiving any feedback on it -- I 
> wouldn't send an additional patch that renames variables. If I imagine 
> that I arrived at this course of action by reason, it might look like 
> this.

Thanks for the (snipped) detailing.

> I thought that sharing what I would do could be useful to xPMo.

Sure, and that's appreciated — but answering "No" without a reason to a proposed contribution is rarely a good idea.  (Over at Apache, for example, -1 votes are expected to always be accompanied by sound technical rationales.)

> It hasn't occurred to me that I may be seen as a reviewer and/or 
> maintainer of zsh. I though I'm commenting on 
> romkatv/powerlevel10k#1042 
> <https://github.com/romkatv/powerlevel10k/issues/1042>. Only now I've 
> realized that this thread is on zsh-users/zsh. I wouldn't write such a 
> short and personal remark as a reviewer/maintainer.

*nod*

> > Also, spelling out a reason can unearth new conclusions.
> 
> True. It's work though.

Even a short explanation (acknowledging its own brevity, if needed) would be better than nothing.

Thanks for straightening this misunderstanding out!

Daniel

PR #71, opened 2021-01-14
update xrandr completion

pseyfert:
Not much magic happening, the output of `xrandr --listproviders` is used to suggest completions after `xrandr --setprovideroutputsource`. No experience from my side about `--setprovideroffloadsink` therefore I left it untouched.

AFAIU `--setprovideroutputsource` accepts multiple arguments (e.g. `xrandr --setprovideroutputsource modesetting  NVIDIA-0`, that's why I added the `*`. Unfortunately that means there is no completion for `--output` and alike anymore once one is after `--setprovideroutputsource`. That doesn't seem ideal to me, happy to hear other suggestions.

### DIFF
--- a/Completion/X/Command/_xrandr
+++ b/Completion/X/Command/_xrandr
@@ -3,10 +3,10 @@
 local curcontext="$curcontext" state line expl
 typeset -A opt_args
 
-__xrandrproviders() {
+__xrandr_providers() {
   local -a provs
-  provs=(${${(f)"$(xrandr --listproviders | tail +2)"}##*:})
-  _describe -t xrandrproviders 'providers' provs
+  provs=(${${(f)"$(xrandr --listproviders)"}[2,-1]##*:})
+  compadd "$@" $provs
 }
 
 _arguments -C \
@@ -53,8 +53,8 @@ _arguments -C \
   "*--addmode:output:->outputs:name" \
   "*--delmode:output:->outputs:name" \
   '--listproviders' \
-  '--setprovideroutputsource:*:provider: __xrandrproviders' \
-  '--setprovideroffloadsink:provider: :sink' \
+  '--setprovideroutputsource:provider:__xrandr_providers:source:__xrandr_providers' \
+  '--setprovideroffloadsink:provider:__xrandr_providers:sink:__xrandr_providers' \
   '--listmonitors' \
   '--listactivemonitors' \
   '--setmonitor:name' \
--- a/Completion/X/Command/_xrandr
+++ b/Completion/X/Command/_xrandr
@@ -1,13 +1,7 @@
 #compdef xrandr
 
-local curcontext="$curcontext" state line expl
+local curcontext="$curcontext" state line expl state_descr
 typeset -A opt_args
-
-__xrandr_providers() {
-  local -a provs
-  provs=(${${(f)"$(xrandr --listproviders)"}[2,-1]##*:})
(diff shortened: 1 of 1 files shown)


phy1729:
Thanks for the patch. As you mention the `*` is non ideal. There's a bit in zshcompsys where optspecs are described that states
> This group may be repeated for options which take
> multiple arguments.  In other words,
> :message1:action1:message2:action2 specifies that
> the option takes two arguments.

Which is how the code currently completes two arguments. So you could `--setprovideroutputsource:provider:__xrandrproviders:source:__xrandrproviders' \`.

In the helper function there's no need to shell out to `tail`, ` ${${${(f)"$(xrandr --listproviders)"}[2,-1]}##*:}` can trim the first line. I'm a little wary of assuming name will be last on the line, but I'm not sure what the correct way to parse it would be. Also for `-t xrandrproviders` I'd lean towards `xrandr-providers` as the convention is to join multi-word tags with hyphens.

From the docs and a bit of searching it seems that `--setprovideroffloadsink` uses provider names as well. I think it'd be fine to update it in a similar fashion as described above.

pseyfert:
Thanks for taking a look and the suggestions.

I now switched away from `_describe` (I anyway don't have item descriptions for the providers). The change also respects the list description from the parent `_arguments call (when one has something like `zstyle ':completion:*:descriptions' format $'%d'` set). A downside is this drops the tag. I didn't spot a completion helper function that takes a tag but no description when quickly scrolling over zshcompsys. Do you have a hint or preference how to handle the list description and tag best? I suspect finding the `-X` in `$@` in the helper function is not the way to go.

Completion/X/Command/_xrandr
```diff
@@ -3,6 +3,12 @@
 local curcontext="$curcontext" state line expl
 typeset -A opt_args
 
+__xrandr_providers() {
+  local -a provs
+  provs=(${${(f)"$(xrandr --listproviders)"}[2,-1]##*:})
+  compadd "$@" $provs
```
danielshahaf, line 9: `compadd "$@" - $provs` or `compadd "$@" -a provs` to avoid option injection.

danielshahaf:
> I suspect finding the `-X` in `$@` in the helper function is not the way to go.

Nope :).  Would `_wanted` with `$state_descr` work?  No time to test it, sorry.

pseyfert:
> > I suspect finding the `-X` in `$@` in the helper function is not the way to go.
> 
> Nope :). Would `_wanted` with `$state_descr` work? No time to test it, sorry.

had to look `state_descr` up, seems to do exactly what i wanted.

phy1729:
A reviewer on IRC noted that the `[2,-1]` acts on the result as a scalar if there are no outputs. I've modified things to handle that and match the surrounding code at https://github.com/phy1729/zsh/commit/07765d590159f180a14af741dbe5639392093f0b . Assuming that it works for you, @pseyfert, I'll commit that.

pseyfert:
Cool thanks to you and the reviewers on irc. Fine with me if you commit the update.

phy1729:
Merged. Thanks for the PR!

(@zsh-users/administrators could I get added to the zsh-committers team so I can close PRs?)

pseyfert:
While I can't help you with closing PRs in general I'll close this here already.

danielshahaf:
I endorse @phy1729's request — he has commit access upstream.

PR #72, opened 2021-01-22
Fix unix command lldb tab-completion

Semnodime:
The --arguments are space (` `) separated, not equal-sign (`=`) separated.
lldb --correct c_arg --wrong=w_arg

### DIFF
--- a/Completion/Unix/Command/_lldb
+++ b/Completion/Unix/Command/_lldb
@@ -6,12 +6,12 @@ typeset -a opts args
 [[ $EUID = 0 || $_comp_priv_prefix[1] = sudo ]] && opts=( -a )
 
 args=(
-  '*'{-o+,--one-line=}'[run one-line lldb command after loading executable]:lldb command: '
-  '*'{-s+,--source=}'[run lldb commands from a file after loading executable]:file:_files'
-  '*'{-O+,--one-line-before-file=}'[run one-line lldb command before loading executable]:lldb command'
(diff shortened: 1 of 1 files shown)


okapia:
Thanks for the contribution, I needed to tweak this slightly as you had changed the equals to a space where space is the default so it needed to be empty. Otherwise from that, It has been merged in ed160f26.

Semnodime:
When will the merge appear as update?

okapia:
I'm not certain that I follow what your question means. If you mean when will there be a 5.9 release of zsh with the change, I can't answer that because I'm not the zsh release manager. We're rather behind on merging patches at the moment so there would be a fair amount of work just to get it in shape. If you're talking about the github pull request merge status, I'm not sure how well that follows because the github repo is just a mirror so isn't used directly.

danielshahaf:
> I'm not certain that I follow what your question means. If you mean when will there be a 5.9 release of zsh with the change, I can't answer that because I'm not the zsh release manager. We're rather behind on merging patches at the moment so there would be a fair amount of work just to get it in shape. If you're talking about the github pull request merge status, I'm not sure how well that follows because the github repo is just a mirror so isn't used directly.

I don't think we should block a 5.9 release on merging all outstanding patches: there's been more than enough patches merged after 5.8 that are awaiting the next train.  Any patches that don't make the 5.9 train will simply have to wait for 5.10.

Anyway, this is a discussion for -workers@.  @Semnodime, feel free to start a thread about scheduling 5.9.

PR #74, opened 2021-03-30
Drop vcsh completions, maintained upstream

alerque:
See discussion on upstream project [here](https://github.com/RichiH/vcsh/issues/270).

The VCSH project is maintaining updated and working completions for ZSH upstream. We'd be happy to also update the ones here but given that the upstream ones are going to be more up to date and match the release cycle for VCSH itself it seems like it might be better to just call those canonical. If it's preferable to keep both I'm happy to send the update through instead.


RichiH:
CC @okapia FYI.

okapia:
Merged in 3afdfef3, Thanks.

PR #75, opened 2021-06-26
vcs_info-examples: optimize +vi-git-untracked()

sunaku:
Speed up the prompt on large and/or deep working directories by
stopping grep(1) as soon as it finds a single match, with `-q`.
Also, optimize the regular expression with an anchor and space.

Previously, +vi-git-untracked() waited for grep(1) to find all
matches of untracked files, redirecting them away to /dev/null,
before finally concluding that untracked files do indeed exist.


sunaku:
Oops, I accidentally created this PR from the zsh-users organization.  I'll close this and recreate it from my own GitHub account.

sunaku:
See PR #76.

PR #76, opened 2021-06-26
vcs_info-examples: optimize +vi-git-untracked()

sunaku:
Speed up the prompt on large and/or deep working directories by
stopping grep(1) as soon as it finds a single match, with `-q`.
Also, correct the regexp by adding a ^ anchor and increase its
specificity by accounting for the space in Porcelain Format v1.

Previously, +vi-git-untracked() waited for grep(1) to find all
matches of untracked files, redirecting them away to /dev/null,
before finally concluding that untracked files do indeed exist.

With this patch, I see 4x speedup on a large Git-enabled $HOME:

    $ time (git status --porcelain | wc -l)
    212
    0.01s user 0.02s system 0% cpu 9.021 total

    $ time (git status --porcelain | grep '??' &>/dev/null)
    0.01s user 0.02s system 0% cpu 12.294 total

    $ time (git status --porcelain | grep -q '^?? ' 2>/dev/null)
    0.01s user 0.01s system 0% cpu 3.097 total

Note that `-q` for grep(1) is in POSIX (IEEE Std 1003.1-2017):
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html

See discussion https://www.zsh.org/mla/workers/2021/msg01354.html

### DIFF
--- a/Misc/vcs_info-examples
+++ b/Misc/vcs_info-examples
@@ -160,7 +160,7 @@ zstyle ':vcs_info:git*+set-message:*' hooks git-untracked
 
 +vi-git-untracked(){
     if [[ $(git rev-parse --is-inside-work-tree 2> /dev/null) == 'true' ]] && \
-        git status --porcelain | grep '??' &> /dev/null ; then
+        git status --porcelain | grep -q '^?? ' 2> /dev/null ; then
         # This will show the marker if there are any untracked files in repo.
         # If instead you want to show the marker only if there are untracked
         # files in $PWD, use:


Misc/vcs_info-examples
```diff
@@ -160,7 +160,7 @@ zstyle ':vcs_info:git*+set-message:*' hooks git-untracked
 
 +vi-git-untracked(){
     if [[ $(git rev-parse --is-inside-work-tree 2> /dev/null) == 'true' ]] && \
-        git status --porcelain | grep '??' &> /dev/null ; then
+        git status --porcelain 2> /dev/null | grep -q '^?? ' ; then
```
danielshahaf, line 163: Why is git's stderr discarded?

Shouldn't the question marks be escaped? (preëxisting issue)

danielshahaf (commented):
Thanks. Adding -q makes sense. Review below.

danielshahaf:
By the way, could you please send a short email to -workers@ pointing to this PR, so interested parties will be able to follow/join the discussion?

Misc/vcs_info-examples
sunaku: I discarded git's stderr because it shouldn't pollute the prompt in case of failure.

Question marks are treated literally by grep's default [BRE syntax](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_03), so this is fine.

sunaku:
> By the way, could you please send a short email to -workers@ pointing to this PR, so interested parties will be able to follow/join the discussion?

Done, please see this thread: https://www.zsh.org/mla/workers/2021/msg01354.html

danielshahaf:
> I discarded git's stderr because it shouldn't pollute the prompt in case of failure.

What about "If you have to fail, don't fail silently"?

Note that other examples only silence stderr if they validate the results afterwards.

> Question marks are treated literally by grep's default BRE syntax 
> <https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_03>, so this is fine.

Ack, thanks.

sunaku:
> I discarded git's stderr because it shouldn't pollute the prompt in case of failure.
> What about "If you have to fail, don't fail silently"? Note that other examples only silence stderr if they validate the results afterwards.

Sure, I'll revert that portion. :sweat_smile:  The original code seemed to me like it wanted to silence the stderr of the entire git+grep pipeline, but in fact, the stderr redirection only applies to grep.  Since grep wouldn't produce any meaningful stderr output in this case, I assumed that the original intent was to silence git. :shrug:

sunaku:
Hi @danielshahaf, I have force-pushed your requested /dev/null reversal and padded the commit message with answers from discussion on the zsh-workers mailing list.  Please review it, at your convenience.  Thank you.

danielshahaf:
Thanks. Looks good to me. Adding the `^` deserves a mention in the log message IMO. I'll merge when I'm back at my workstation. Excuse brevity.

danielshahaf:
Sorry, adding the anchor is already mentioned (I did look before my previous comment), but it's described as an "optimization", which isn't accurate: the code changes improves not only performance but also correctness. The log message should mention that.

sunaku:
How about this?
> Also, correct the regex with ^ anchor and increase specificity.

danielshahaf:
s/with/by adding a/; s/specificity/its specificity/; then +1. Thanks!

sunaku:
Done, with a cherry on top! :cherries: :ice_cream: :yum: 

>Also, correct the regexp by adding a ^ anchor and increase its
>specificity by accounting for the space in Porcelain Format v1.

danielshahaf:
Thanks a lot! Will merge from my workstation. :-)

danielshahaf:
Merged in d634a3d33302.  Thanks a lot again!

PR #77, opened 2021-07-27
Improve `subl` completions for Sublime Text v4

stdedos:
Additionally, add model lines

Signed-off-by: Stavros Ntentos <133706+stdedos@users.noreply.github.com>

### DIFF
--- a/Completion/X/Command/_sublimetext
+++ b/Completion/X/Command/_sublimetext
@@ -8,6 +8,8 @@ _arguments -s -S : \
   '(-a -n --add --new-window)'{-n,--new-window}'[open new window]' \
   '--project[load specified project]:project file:_files' \
   '(-s --stay)'{-s,--stay}'[keep application activated]' \
+  '--safe-mode[launch using sandboxed (clean) environment]' \
+  '--launch-or-new-window[only open new window if application is open]' \
   '(: -)'{-v,--version}'[display version information]' \
   '(-w --wait)'{-w,--wait}'[wait for files to close]' \
   '*:file or directory:_files'


Completion/X/Command/_sublimetext
```diff
@@ -7,7 +7,16 @@ _arguments -s -S : \
   '(: -)'{-h,--help}'[display help information]' \
   '(-a -n --add --new-window)'{-n,--new-window}'[open new window]' \
   '--project[load specified project]:project file:_files' \
-  '(-s --stay)'{-s,--stay}'[keep application activated]' \
```
phy1729, line 10: [Sublime's docs][1] contain "The available flags will vary per operating system," so I'm hesitant to remove any flags without an ok from @okdana (who added the flag).

[1]: https://www.sublimetext.com/docs/command_line.html

Completion/X/Command/_sublimetext
```diff
@@ -7,7 +7,16 @@ _arguments -s -S : \
...
+  '(--safe-mode)'--safe-mode'[Launch using a sandboxed (clean) environment]' \
```
phy1729, line 10: The other specs are in the form `'(-x --xlong-form)'{-x.--xlong-form}'[description]'` because there is a short and long form for the same setting. Since `--safe-mode` only has a long option, `'--safe-mode[Launch using a sandboxed (clean) environment]'` suffices. Same for `--launch-or-new-window`.

Completion/X/Command/_sublimetext
```diff
@@ -7,7 +7,16 @@ _arguments -s -S : \
...
   '(-w --wait)'{-w,--wait}'[wait for files to close]' \
   '*:file or directory:_files'
+
+# Local Variables:
+# mode: Shell-Script
+# sh-indentation: 2
+# indent-tabs-mode: nil
+# sh-basic-offset: 2
+# End:
+# vim: ft=zsh sw=2 ts=2 et
```
phy1729, line 22: As most completers don't have mode lines, I think this part should be left out.

Completion/X/Command/_sublimetext
stdedos: Would've been nice to also conditionally output flag - but at least Sublime does not seem to do anything either way on Ubuntu.

I can revert since you hesitate.

Completion/X/Command/_sublimetext
stdedos: I have no idea how the syntax works (and https://github.com/vapniks/zsh-completions/blob/master/zsh-completions-howto.org, https://zsh.sourceforge.io/Doc/Release/Completion-System.html#:~:text=grouping%20options don't seem to explain it adequately to my level of understanding).

Completion/X/Command/_sublimetext
danielshahaf: Conditional behaviour is possible. What would the condition be?

Completion/X/Command/_sublimetext
danielshahaf: In the second link, search for "Each of the forms above may be preceded by a list in parentheses of option names and argument numbers". The grouping facility is unrelated and rarely used.

Completion/X/Command/_sublimetext
stdedos: All arguments must exist in `subl --help` (by means of grepping??), otherwise not offered

I know it's not a panacea convention, but a lot of respectful utils offer their arguments on either `-h`/`--help` - Sublime does too

okdana: `-s`/`--stay` is still there on macOS at least in `subl` 4113. If that option is platform-specific, like Daniel said, we could easily make it conditional. Determining the applicability of options by looking at `--help` output is possible, but not common; in this case  i think it'd be fine to just look at `$OSTYPE`

Completion/X/Command/_sublimetext
```diff
@@ -8,6 +8,8 @@ _arguments -s -S : \
   '(-a -n --add --new-window)'{-n,--new-window}'[open new window]' \
   '--project[load specified project]:project file:_files' \
   '(-s --stay)'{-s,--stay}'[keep application activated]' \
+  '--safe-mode[Launch using a sandboxed (clean) environment]' \
```
okdana, line 11: Completion descriptions should not use sentence case (they should start with a lower-case letter). And since the existing descriptions use 'telegraph' style i would make these match (omit 'a' and 'the' where possible)

okdana (changes requested):

Completion/X/Command/_sublimetext
stdedos: I disagree, but whatever - if those are the rules ...

Completion/X/Command/_sublimetext
stdedos: > in this case i think it'd be fine to just look at `$OSTYPE`

"As far as you know" it's based on `$OSTYPE`. Maybe it's only on MacOS? Or it's only not on Linux? 😕
Blackbox approach (e.g. grepping) makes less assumptions on things we don't know for sure.

However, I am not going to tackle it anyway - especially now that you verified it's still a thing.

Completion/X/Command/_sublimetext
okdana: The no-sentence-case thing is mentioned in [the style guide](https://github.com/zsh-users/zsh/blob/master/Etc/completion-style-guide) (under *Descriptions*), and most other functions shipped with zsh follow it. I'm not sure if we have any specific guidance re: telegraph style, but it's certainly common, and if nothing else it's usually a good idea to match a file's existing style

stdedos: > The no-sentence-case thing is mentioned in the style guide (under Descriptions), and most other functions shipped with zsh follow it. ...

Yeah, I got that - hence the new commit.

However, it sounds so ... weird.
And technically, also on the same page it mentions:
> It is a good idea to copy descriptions from the command's man page or --help output.

Completion/X/Command/_sublimetext
okdana: > Maybe it's only on MacOS? Or it's only not on Linux?

Either/both of those conditions could be handled by checking `$OSTYPE`, and it's how most other functions with platform-specific options do it. But yeah i haven't actually checked to see which options are available on which platforms

> However, I am not going to tackle it anyway

That's fine, we can revisit later if we want

okdana (approved):
Looks OK, thank you. I will merge later (unless someone else wants to do it)

Completion/X/Command/_sublimetext
danielshahaf: That just advises not to reinvent the wheel when describing options. Some degree of editing is usually necessary, e.g., when the --help output of an option that takes an argument names the argument.

okdana:
Sorry for the delay. Merged in 765bc1470

PR #79, opened 2021-10-15
update options in apple's version of ssh-add

thelastlin:
ssh-add behavior has changed in newer macOS releases.
```
$ APPLE_SSH_ADD_BEHAVIOR= ssh-add -K
WARNING: The -K and -A flags are deprecated and have been replaced
         by the --apple-use-keychain and --apple-load-keychain
         flags, respectively.  To suppress this warning, set the
         environment variable APPLE_SSH_ADD_BEHAVIOR as described in
         the ssh-add(1) manual page.
...

$ APPLE_SSH_ADD_BEHAVIOR=openssh ssh-add -A
ssh-add: illegal option -- A
...

$ APPLE_SSH_ADD_BEHAVIOR=macos ssh-add -A
Identity Added: /path/to/key (keys comment)
...
```

### DIFF
--- a/Completion/Unix/Command/_ssh
+++ b/Completion/Unix/Command/_ssh
@@ -81,18 +81,20 @@ _ssh () {
       '*:file:->file' "$common[@]" "$common_transfer[@]" && ret=0
     ;;
   ssh-add)
-    args=(
-      '-K[load resident keys from a FIDO authenticator]'
-    )
-    if [[ $OSTYPE == darwin* ]]; then 
-      [[ $APPLE_SSH_ADD_BEHAVIOR == "macos" ]] && args=(
+    if [[ $OSTYPE != darwin* || $APPLE_SSH_ADD_BEHAVIOR == openssh ]]; then
+      args=(
+        '-K[load resident keys from a FIDO authenticator]'
+      )
+    else
+      [[ ${APPLE_SSH_ADD_BEHAVIOR:-macos} == macos ]] && args=(
         '-A[add identities from keychain]'
         '-K[update keychain when adding/removing identities]'
-      ) || args+=(
-        '--apple-load-keychain[add identities from keychain]'
-        '--apple-use-keychain[update keychain when adding/removing identities]'
       )
-   fi
+    fi
+    [[ $OSTYPE == darwin* ]] && args+=(
+      '--apple-load-keychain[add identities from keychain]'
+      '--apple-use-keychain[update keychain when adding/removing identities]'
+    )
     _arguments -s : $args \
       '-c[identity is subject to confirmation via SSH_ASKPASS]' \
       '-D[delete all identities]' \
--- a/Completion/Unix/Command/_ssh
+++ b/Completion/Unix/Command/_ssh
@@ -81,10 +81,18 @@ _ssh () {
       '*:file:->file' "$common[@]" "$common_transfer[@]" && ret=0
     ;;
   ssh-add)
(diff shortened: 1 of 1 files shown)


phy1729:
I can't reproduce this on macOS 11.6. If this is a change in 12, we should probably support both versions for a while (not sure what our policy is here if one exists).

thelastlin:
> I'm still on Catalina, but i found [one result](https://keith.github.io/xcode-man-pages/ssh-add.1.html#APPLE_SSH_ADD_BEHAVIOR) for this on Google, which is apparently taken from the Xcode 13 beta and says:
> 
> > APPLE_SSH_ADD_BEHAVIOR
> > Enables or disables the older processing of the -A and -K options used in earlier macOS releases. These options have been renamed --apple-load-keychain and --apple-use-keychain respectively. However, -A and -K still behave as in earlier releases except in the following circumstances. If a security provider was specified with -S or SSH_SK_PROVIDER, or if APPLE_SSH_ADD_BEHAVIOR is set to the value “openssh”, then ssh-add uses standard OpenSSH behavior: the -A flag is not recognized and the -K flag behaves as documented in the DESCRIPTION section above.
> > Otherwise, ssh-add -A and -K will behave as in earlier macOS releases. A warning will be output to standard error unless APPLE_SSH_ADD_BEHAVIOR is set to the value “macos”. Note: Future releases of macOS will not support neither -A nor -K without setting this environment variable.
> 
> Our usual policy is to update completion functions for whatever's in the latest version of the tool and let people with older versions just deal with that. In this case, if the above is accurate, for now `-A` and `-K` still function as before by default, so i don't think we need to worry about it either way, just rework the change a little to match the documentation better. Maybe something like this?
> 
> ```
> if
>   [[ $OSTYPE != darwin* ]] ||
>   [[ $APPLE_SSH_ADD_BEHAVIOR == openssh ]] ||
>   [[ -n $SSH_SK_PROVIDER ]]
> then
>   args=(
>     '-K[load resident keys from a FIDO authenticator]'
>   )
> elif [[ ${APPLE_SSH_ADD_BEHAVIOR:-macos} == macos ]]; then
>   args=(
>     '-A[add identities from keychain]'
>     '-K[update keychain when adding/removing identities]'
>   )
> fi
> 
> [[ $OSTYPE == darwin* ]] && args+=(
>   '--apple-load-keychain[add identities from keychain]'
>   '--apple-use-keychain[update keychain when adding/removing identities]'
> )
> ```
> 
> In a future update we could remove the `:-macos` fall-back from the second condition. Also, if we wanted to be fancier we could make the last condition check for like `[[ $OSTYPE == darwin<20->.* ]]`, though that's not super accurate since `OSTYPE` is set at compile time. Also this doesn't handle the `-S` scenario the man page describes, only the corresponding environment variable; if that distinction is confusing we could leave out the `SSH_SK_PROVIDER` check

Thanks for your advice. Changes is applied in https://github.com/zsh-users/zsh/pull/79/commits/a3e80569785415fd8eefee0b490fad4eb25fca0a .

okdana:
@phy1729 told me that he prefers the new long options to be version-gated. I don't think it really matters personally (i'm sure Catalina's install base has already shrunk to almost nothing by now), but apparently Homebrew's zsh has an accurate `OSTYPE` for Big Sur, so there shouldn't be much harm in it.

I've committed the change in 0f1e4d8e7 with that minor adjustment. Hope that's OK with you. Thanks!

PR #80, opened 2021-11-04
_globquals: invert before/since for date glob qualifiers completion

vincentbernat:
The documentation says:

> Files accessed within the last n days are selected using a negative
> value for n (-n). Files accessed more than n days ago are selected
> by a positive n value (+n).

So, negative values are files since the selected date and positive
values are files before the selected days.

### DIFF
--- a/Completion/Zsh/Type/_globquals
+++ b/Completion/Zsh/Type/_globquals
@@ -127,15 +127,15 @@ while [[ -n $PREFIX ]]; do
 	alts+=( "time-specifiers:time specifier:compadd -E 0 -d tdisp -S '' -a tmatch" )
       fi
       if ! compset -P '[-+]' && [[ -z $PREFIX ]]; then
-	sdisp=( before exactly since )
+	sdisp=( since exactly before )
 	smatch=( - '' + )
 	if zstyle -t ":completion:${curcontext}:senses" verbose; then
 	  zstyle -s ":completion:${curcontext}:senses" list-separator sep || sep=--
 	  default=" [default exactly]"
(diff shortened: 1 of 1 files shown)


okapia:
This particular thing was already toggled once before in the opposite direction. git blame points to me but I'm fairly certain someone else told be to swap it and the original sense had been according to my intuition.

It depends whether you read it as the file time is before the specified time or the other way around. Maybe we need to rethink how this is presented entirely because different people seemingly have different intuition in this regard as to which is logical.

vincentbernat:
`echo *(m` then `<tab>` provides `before` and `since` for `sense`, then a calendar view. So, if I read aloud, I would say "all files modified before December 12" and I will get `*(m-2)` which will select files modified *after* December 12. How do you "read aloud" the completion widget in your case?

danielshahaf:
> Maybe we need to rethink how this is presented entirely because different people seemingly have different intuition in this regard as to which is logical.

How about adding a parenthetical? —

```
- -- since (match files newer than)
+ -- before (match files older than)
```

okapia:
I'll post an updated patch to the mailing list. Partly to preserve the ordering of before coming ahead of since in the completion list. I'll go with a shorter parenthetical, just (older files) and (newer files). Briefer descriptions are easier to scan and I want to avoid the completion wrapping to a new line. The calendar completion works best when plenty of the terminal is available to it. The wording was "more than" and "less than" at one point.

danielshahaf:
> I'll post an updated patch to the mailing list.

Posted https://zsh.org/workers/49648.  See y'all there :)

phy1729:
Incorporated into 702d773.

PR #83, opened 2021-11-24
[completion] git push -o can be set multiple times.

pseyfert:
From [the git-push docs](https://git-scm.com/docs/git-push#Documentation/git-push.txt---push-optionltoptiongt) I understand that `-o` can be provided multiple times. This also matches the usage examples [by gitlab](https://docs.gitlab.com/ee/user/project/push_options.html).


danielshahaf:
LGTM. Does `_git-send-pack` need the same change?

pseyfert:
Didn't find it in the docs, but looking at [the source](https://github.com/git/git/blob/master/builtin/send-pack.c#L199) you're right.

danielshahaf:
Pushed in 5fe498124d6469fadded7b9ad9eb64649b93f2de.  Thanks!

PR #85, opened 2021-12-26
replace every occurrence of replace_string

EdwarDu:
`zargs -i{} -- hi -- echo "{} and {} is okay"` will output 'hi and {} is okay'
with the change will output 'hi and hi is okay'

### DIFF
--- a/Functions/Misc/zargs
+++ b/Functions/Misc/zargs
@@ -291,7 +291,7 @@ do
 	args=( "${(@)argv[1,end]}" )
 	shift $((end > ARGC ? ARGC : end))
 	if (( $#i ))
-	then call=( "${(@)command/$i/$args}" )
+	then call=( "${(@)command//$i/$args}" )
 	else call=( "${(@)command}" "${(@)args}" )
 	fi
 	if (( ${(c)#call} > s ))


larryv:
Thanks for the suggestion, but [we only accept GitHub PRs for completion functions](https://zsh.sourceforge.io/Arc/git.html).  Please submit the patch to <zsh-workers@zsh.org> for review. (You don't have to subscribe.)

danielshahaf:
> [we only accept GitHub PRs for completion functions](https://zsh.sourceforge.io/Arc/git.html).

We should state this in zsh.git too, because people may get the source code without going through <https://zsh.sourceforge.io/Arc/git.html>.  Do we?  There are no relevant mentions of "github", but there may be a reference to "see the Web site for instructions" or some such, which isn't easy to grep for.

EdwarDu:
> Thanks for the suggestion, but [we only accept GitHub PRs for completion functions](https://zsh.sourceforge.io/Arc/git.html). Please submit the patch to [zsh-workers@zsh.org](mailto:zsh-workers@zsh.org) for review. (You don't have to subscribe.)

Thank you for the info. I will send out the email later. Would you mind let me know if there is specific format for the "patch" email ?

larryv:
> Would you mind let me know if there is specific format for the "patch" email ?

Assuming your email client can send plain text messages without mangling them (e.g., adding unwanted newlines), then inline `git format-patch` or `git diff` output works well. ([An example](https://www.zsh.org/mla/workers/2021/msg01894.html). [Another example](https://www.zsh.org/mla/workers/2021/msg01893.html).)

If your email client does mangle messages (the Gmail web client seems particularly troublesome), or if you're not sure whether it does, then an attachment with a `.txt` extension is best.

larryv:
> We should state this in zsh.git too, because people may get the source code without going through <https://zsh.sourceforge.io/Arc/git.html>. Do we? There are no relevant mentions of "github", but there may be a reference to "see the Web site for instructions" or some such, which isn't easy to grep for.

I skimmed [Etc/zsh-development-guide](../blob/master/Etc/zsh-development-guide) and didn't see anything.  I don't think it acknowledges the mirrors' existence at all.

(That file is already not particularly discoverable; I can't imagine many people come across it, let alone read it.)

okapia:
@larryv if you identify things like this, patches against the text files are always welcome. a reference to Etc/zsh-development-guide and Etc/completion-style-guide in the Documentation section of the README would make sense.

A much broader reorganisation of the text files is somewhat needed. E.g. consolidating the various bits that should constitute release notes. And perhaps using markdown.

It's often difficult to get the balance of including all the relevant information where it can be found and there not being so much that nobody can be bothered to read it. That's why the web page only briefly mentions that PRs for Completion functions are accepted without going into the details of why. For a PR against zargs it appears finicky to ask for a patch but I don't use zargs so can't give meaningful consideration to the change.

danielshahaf:
> For a PR against zargs it appears finicky to ask for a patch but I don't use zargs so can't give meaningful consideration to the change.

Project policy is that patches to anything other than Completion/ are to go through the mailing list.  So long as that's the policy, people interested in zargs might not be subscribed to github/gitlab or the push notifications stream, but only to -workers@.  Accepting zargs patches via github would break those people's reliance on the documented project policy.  Feel free to drive a policy change.

I'm not expressing an opinion on such a change.

EdwarDu:
Thank you all for the input, I'm closing this now.

PR #87, opened 2022-02-18
[completion] Update _tmux definition using a recent build of tmux

atomicstack:
Using the tool [`zsh.git:Util/check-tmux-state`](https://github.com/zsh-users/zsh/blob/master/Util/check-tmux-state), I have generated some updates for [tmux completion in zsh](https://github.com/zsh-users/zsh/blob/master/Completion/Unix/Command/_tmux). This also required some minor tweaks to the tool itself.

The update was based off tmux.git revision [c67abcf8182b](https://github.com/tmux/tmux/commit/c67abcf8182b3a4e4c1e71b370c814a65c12a46c).

Most updates were simple to apply, however, there is some oddness related to the `server-info` and `info` commands due to changes in tmux, in order to enable dynamic aliases. As of tmux.git commit [126d364abe2e8c063efa5a888de69](https://github.com/tmux/tmux/commit/126d364abe2e8c063efa5a888de6997ca05641fc), `server-info` is no longer a command, and both `server-info` and its existing alias `info` are now defined as a `command-alias` for `show-messages -JT`.

`check-tmux-state` assumes that aliases are listed in the output of `tmux list-commands` and nowhere else, and so [creates `$available_aliases` by looking for alias names enclosed in parens](https://github.com/zsh-users/zsh/blob/master/Util/check-tmux-state#L70):
```
% grep -A2 'available_aliases=' Util/check-tmux-state
available_aliases=( $( $tmux list-commands |
                       grep '^[a-z-]* *(' |
                       gsed -e 's,^\([a-z-]*\) *(\([a-z-]*\))\(.*\)$,\2 \1,' ) )
# splitw -> split-window
% tmux list-commands| grep ^split
split-window (splitw) [-bdefhIPvZ] [-c start-directory] [-e environment] [-F format] [-l size] [-t target-pane][shell-command]
```

But, it's now also necessary to update `$available_aliases` based on the `command-alias` entries in tmux's server options:
```
# dynamic alias config:
% tmux show-options -s command-alias | egrep info
command-alias[2] "server-info=show-messages -JT"
command-alias[3] "info=show-messages -JT"
```

...and this can't be done (easily) by using the existing code, as it's not possible for zsh to use a hyphenated token like `server-info` as a key in an associative array AFAICT. Attempting to add such an item results either in a silent error, or alternatively, a bad subscript error.

This will also affect the usage of `zstyle`. The `_tmux` completion file suggests that completion output for tmux can be restricted to commands only (or aliases only), with a line such as the following:
```
#       % zstyle ':completion:*:*:tmux:*:subcommands' mode 'commands'
```
If a user chooses to use this technique to enable commands only, then important aliases such as server-info will no longer be visible to the user.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2022-02-18  Matt Koscica <matt.koscica@gmail.com>
+
+	* github PR #87: Completion/Unix/Command/_tmux: update options
+	for tmux 3.4
+
 2022-02-16  Jun-ichi Takimoto  <takimoto-j@kba.biglobe.ne.jp>
 
 	* 49757 (sourceforge #1): samcarter: Completion/Unix/Command/_tex:
--- a/Completion/Unix/Command/_tmux
+++ b/Completion/Unix/Command/_tmux
@@ -130,6 +130,8 @@ _tmux_aliasmap=(
     popup       display-popup
 
     # buffers
+    clearphist  clear-prompt-history
+    showphist   show-prompt-history
     clearhist   clear-history
     deleteb     delete-buffer
     lsb         list-buffers
@@ -267,6 +269,11 @@ _tmux-clear-history() {
   _arguments '-t+[specify target pane]:pane:__tmux-panes'
 }
 
+_tmux-clear-prompt-history() {
+  [[ -n ${tmux_describe} ]] && print "remove history of values for command line prompts" && return
+  _arguments '-T+[specify prompt type]'
+}
+
 _tmux-clock-mode() {
   [[ -n ${tmux_describe} ]] && print "enter clock mode" && return
   _arguments '-t+[specify target pane]:pane:__tmux-panes'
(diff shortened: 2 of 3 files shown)


phy1729:
Looks generally ok. I know the cut was already there, but the cut, sed, and uniq can be done with zsh features. That is
```zsh
typeset -a available_session_options
available_session_options=( $( $tmux show-options -g | cut -f1 -d' ' | sed 's/\[[0-9+]\]$//' | uniq ) )
```
could instead be
```zsh
typeset -aU available_session_options
available_session_options=( ${${(f)"$(tmux show-options -g)"}%%(\[<->\])# *} )
```

atomicstack:
The dependencies upon `cut`+`sed` are gone. However, I had to introduce a dependency upon `grep` to exclude user-defined options (which are prefixed with an `@` symbol), which seemed conceptually cleaner than updating `find_new`, `find_old`, and `find_changed_scope` to ignore that prefix. Definitely open to a better suggestion here.

danielshahaf:
> However, I had to introduce a dependency upon `grep` to exclude user-defined options (which are prefixed with an `@` symbol), which seemed conceptually cleaner than updating `find_new`, `find_old`, and `find_changed_scope` to ignore that prefix. Definitely open to a better suggestion here.

Use `${haystack:#pattern}` instead of `grep -v`.  Calling tmux(1) with `-f /dev/null` might also do the trick.

```zsh
% set -- foo bar baz 
% print -raC1 -- ${argv:#b*} 
foo
% 
```

danielshahaf:
Feel free to post this to zsh-workers@zsh.org; you may get a faster response there.

There's no need to include ChangeLog hunks in patches (but thanks for doing so).

okapia:
merged in dbefe08f. Thanks for the contribution.

Tweaked it as per Daniel's suggestion to use :# instead of grep -v and also to fix a couple of descriptions to use the conventional form.

PR #89, opened 2022-04-09
fix and extend htop completions

xpufx:
I noticed htop was suggesting -v for --version but it should be -V .  (Checked on alpine, fedora, nixos and ubuntu)
Also added some missing options based on htop --help .

I actually don't know how to write completion files but tested the changes. Feel free to make any changes or ignore the pull request and copy to master directly.

### DIFF
--- a/Completion/Linux/Command/_htop
+++ b/Completion/Linux/Command/_htop
@@ -1,11 +1,15 @@
 #compdef htop
 
 _arguments -S : \
-  '(-d --delay)'{-d+,--delay=}'[update frequency]:duration (tenths of seconds)' \
   '(-C --no-color --no-colour)'{-C,--no-colo{,u}r}'[monochrome mode]' \
+  '(-d --delay)'{-d+,--delay=}'[Set the delay between updates]:duration (tenths of seconds)' \
+  '(-F --filter)'{-F+,--Filter=}'[Show only the commands matching the given filter]' \
   '(-)'{-h,--help}'[display usage information]' \
+  '(M --no-mouse)'{-M,--no-mouse}'[Disable the mouse]' \
   \*{-p+,--pid=}'[show given pids]: : _sequence -n ${$(</proc/sys/kernel/pid_max)\:-32768} _pids' \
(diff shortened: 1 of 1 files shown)


okdana:
Some comments on this function (not necessarily your changes specifically):

* Descriptions shouldn't use sentence case (don't capitalise the first letter)

* The wording given in the command's help/manual isn't necessarily what's best or conventional for completion. e.g. 'display version information' is the usual wording we use for `--version`, and i think we generally prefer to omit unnecessary articles and stuff

* The `--delay` description could make use of Oliver's recent updates re: units/ranges/defaults. The default isn't mentioned in the manual but it seems to be 15 (1.5 seconds)

* `--filter` should be lower-case, and maybe we want to complete something there? I think it's just a case-insensitive sub-string search, so it could be anything, but it might be nice to complete command names or path components. At the very least we can add a more helpful description

* The PID limit on `_sequence` is kind of weird. It's probably wrong on most non-Linux systems (i know it is on macOS), and even where it isn't, the user would probably hit `ARG_MAX` before they could reach the PID limit, and even where they wouldn't, are we really that worried that a user will complete tens of thousands of PIDs lol

* `--sort-key` completion is broken on htop 3.1.2 (works on 2.2.0)

* There are a few other options missing from here, like `--highlight-changes` and the Linux-only `--drop-capabilities`

I may use this patch as a basis for a larger update (would credit you obv)

okdana:
np, your change is perfectly fine aside from those very minor wording issues

okdana:
Sorry for the delay. I expected the 5.9 release to move a little more swiftly so i procrastinated on this

I've posted my improved patch to the ML: https://www.zsh.org/mla/workers/2022/msg00506.html

This change might be too big to go in before 5.9, but i'm double-checking on that. Feel free to test or provide feedback in the mean time, if you like

PS: Since you did the initial work on this i intend to commit it in your name. Please let me know if you object

xpufx:
@okdana You've done most of the work. Please commit in your name.

okdana:
I've pushed my change in eec9882d04661b1114e93b6bf6129453fae6f5a1, it will ship with zsh 5.9

I committed it in my name as you requested. Thanks for the initial work you did on it, anyway. Please let us know if you find any other problems with it

PR #91, opened 2022-05-13
Fix issues with find completion

VorpalBlade:
The following issues existed and were fixed:
* -exec and -execdir can take ; or + as an ending marker. Previously only ; was
  supported. This is part of POSIX for -exec (support for + for -ok is optional
  and none of the implementations I looked at seem to support that).
* Missing completion for -files0-from (GNU find 4.9.0 and later). This flag
  needs to go with the global flags at the beginning and load the paths from
  the given file instead of from the command line.
* Missing completion for the -newerXY family of flags (GNU find 4.3.3, also
  available in FreeBSD and MacOS at least).

### DIFF
--- a/Completion/Unix/Command/_find
+++ b/Completion/Unix/Command/_find
@@ -37,6 +37,8 @@ case $variant in
     args+=(
       '*-Bmin:birth time (minutes)'
       '*-Bnewer:file to compare (birth time):_files'
+      '*-newer'{a,B,c,m}{a,B,c,m}'[succeed if timestamp 1 is newer than timestamp 2 of the given file]:reference file:_files'
+      '*-newer'{a,B,c,m}t'[succeed if timestamp of given file is newer than timestamp]:timestamp: '
       '*-Btime:birth time (hours)'
     )
   ;|
@@ -46,7 +48,7 @@ case $variant in
       '*-anewer:file to compare (access time):_files'
       '*-cnewer:file to compare (inode change time):_files'
       '*-empty'
-      '*-execdir:program: _command_names -e:*\;::program arguments: _normal'
+      '*-execdir:program: _command_names -e:*(\;|+)::program arguments: _normal'
       '*-maxdepth:maximum search depth'
       '*-mindepth:minimum search depth'
       '*-path:path pattern to search:'
@@ -110,6 +112,7 @@ case $variant in
       '*-readable'
       '*-writable'
       '*-xtype:file type:((b\:block\ special\ file c\:character\ special\ file d\:directory p\:named\ pipe f\:normal\ file l\:symbolic\ link s\:socket))'
+      '-files0-from[search NUL separated paths from FILE]:file:_path'
       '*-fls:output file:_files'
       '*-fprint:output file:_files'
       '*-fprint0:output file:_files'
@@ -126,7 +129,7 @@ _arguments -C $args \
   '*-atime:access time (days):->times' \
   '*-ctime:inode change time (days):->times' \
   '*-depth' \
-  '*-exec:program: _command_names -e:*\;::program arguments: _normal' \
+  '*-exec:program: _command_names -e:*(\;|+)::program arguments: _normal' \
   '*-follow' \
   '*-fstype:file system type:_file_systems' \
   '*-group:group:_groups' \
@@ -147,7 +150,7 @@ _arguments -C $args \
   '*-user:user:_users' \
   '*-xdev' \
   '*-a' '*-o' \
-  '(-D -E -H -L -O -P -f -s -x --help --version)*:directory:_files -/' \
+  '(-D -E -H -L -O -P -f -s -x -files0-from --help --version)*:directory:_files -/' \
 && ret=0
 
 if [[ $state = times ]]; then
--- a/Completion/Unix/Command/_find
+++ b/Completion/Unix/Command/_find
@@ -38,7 +38,10 @@ case $variant in
       '*-Bmin:birth time (minutes)'
       '*-Bnewer:file to compare (birth time):_files'
       '*-newer'{a,B,c,m}{a,B,c,m}'[if [aBcm\]time is newer than [aBcm\]time of given file]:reference file:_files'
-      '*-newer'{a,B,c,m}t'[succeed if timestamp of given file is newer than timestamp]:timestamp: '
+      '*-newerat[if access time is newer than given timestamp]:timestamp: '
+      '*-newerBt[if birth time is newer than given timestamp]:timestamp: '
+      '*-newerct[if creation time is newer than given timestamp]:timestamp: '
+      '*-newermt[if modification time is newer than given timestamp]:timestamp: '
       '*-Btime:birth time (hours)'
     )
   ;|
--- a/Completion/Unix/Command/_find
+++ b/Completion/Unix/Command/_find
@@ -37,7 +37,7 @@ case $variant in
     args+=(
       '*-Bmin:birth time (minutes)'
       '*-Bnewer:file to compare (birth time):_files'
-      '*-newer'{a,B,c,m}{a,B,c,m}'[succeed if timestamp 1 is newer than timestamp 2 of the given file]:reference file:_files'
+      '*-newer'{a,B,c,m}{a,B,c,m}'[if [aBcm\]time is newer than [aBcm\]time of given file]:reference file:_files'
       '*-newer'{a,B,c,m}t'[succeed if timestamp of given file is newer than timestamp]:timestamp: '
       '*-Btime:birth time (hours)'
     )
@@ -112,7 +112,7 @@ case $variant in
       '*-readable'
       '*-writable'
       '*-xtype:file type:((b\:block\ special\ file c\:character\ special\ file d\:directory p\:named\ pipe f\:normal\ file l\:symbolic\ link s\:socket))'
-      '-files0-from[search NUL separated paths from FILE]:file:_path'
+      '-files0-from[start recursing from targets in given file]:NUL-separated targets file:_files'
       '*-fls:output file:_files'
       '*-fprint:output file:_files'
       '*-fprint0:output file:_files'
@@ -150,7 +150,7 @@ _arguments -C $args \
   '*-user:user:_users' \
   '*-xdev' \
   '*-a' '*-o' \
-  '(-D -E -H -L -O -P -f -s -x -files0-from --help --version)*:directory:_files -/' \
+  '(-D -E -H -L -O -P -f -s -x --files0-from --help --version)*:directory:_files -/' \
 && ret=0
 
 if [[ $state = times ]]; then


VorpalBlade:
Note! I only have local access to GNU find. For the other platforms it is based on reading man pages. I have verified against man pages locally for GNU and man pages online for FreeBSD, OpenBSD and MacOS. I have not looked at other platforms as there are too many to go through, and I'm not sure if some of them are even still around.

Completion/Unix/Command/_find
```diff
@@ -126,7 +129,7 @@ _arguments -C $args \
   '*-atime:access time (days):->times' \
   '*-ctime:inode change time (days):->times' \
   '*-depth' \
-  '*-exec:program: _command_names -e:*\;::program arguments: _normal' \
+  '*-exec:program: _command_names -e:*(\;|+)::program arguments: _normal' \
```
danielshahaf, line 132: This part does seem to have the effect of causing `find . -exec ls \; -<TAB>` to offer find(1) options rather than ls(1) options, but I can't get it to _offer_ `\;` or `+` as a completion, and in any case it seems to be relying on undocumented behaviour of `_arguments`.  (That's a preëxisting issue.)

For reference, this line is an `_arguments` optspec for the `-exec` option, with a leading `*` meaning the option can repeated, and with completion for multiple following words (see "This group may be repeated" in the documentation of `_arguments`).  The part `*\;:` is supposed to be another optspec, but neither `\;` nor `(\;|+)` is a valid optspec.

Completion/Unix/Command/_find
```diff
@@ -37,6 +37,8 @@ case $variant in
     args+=(
       '*-Bmin:birth time (minutes)'
       '*-Bnewer:file to compare (birth time):_files'
+      '*-newer'{a,B,c,m}{a,B,c,m}'[succeed if timestamp 1 is newer than timestamp 2 of the given file]:reference file:_files'
```
danielshahaf, line 40: It's not clear to me what "timestamp 1" and "timestamp 2" are.  Instead, perhaps:

```suggestion
      '*-newer'{a,B,c,m}{a,B,c,m}'[if [aBcm\]time is newer than [aBcm\]time of given file]:reference file:_files'
```

Could probably do even better than this.

Completion/Unix/Command/_find
```diff
@@ -37,6 +37,8 @@ case $variant in
...
+      '*-newer'{a,B,c,m}t'[succeed if timestamp of given file is newer than timestamp]:timestamp: '
```
danielshahaf, line 41: "given" is what's in find(1)'s argv, and due to limited horizontal space I'd like to shorten "succeed if" which would otherwise be on almost every line, so:

```suggestion
      '*-newerat[if access time is newer than given timestamp]:timestamp: '
      '*-newerBt[if birth time is newer than given timestamp]:timestamp: '
      '*-newerct[if creation time is newer than given timestamp]:timestamp: '
      '*-newermt[if modification time is newer than given timestamp]:timestamp: '
```

Completion/Unix/Command/_find
```diff
@@ -110,6 +112,7 @@ case $variant in
       '*-readable'
       '*-writable'
       '*-xtype:file type:((b\:block\ special\ file c\:character\ special\ file d\:directory p\:named\ pipe f\:normal\ file l\:symbolic\ link s\:socket))'
+      '-files0-from[search NUL separated paths from FILE]:file:_path'
```
danielshahaf, line 115: The description can't say FILE since the list of options doesn't say `-files0-from FILE`.

`_path` doesn't exist.  Should be `_files`, probably.

How about:

```suggestion
      '-files0-from[start recursing from targets in given file]:NUL-separated targets file:_files'
```

Completion/Unix/Command/_find
```diff
@@ -46,7 +48,7 @@ case $variant in
       '*-anewer:file to compare (access time):_files'
       '*-cnewer:file to compare (inode change time):_files'
       '*-empty'
-      '*-execdir:program: _command_names -e:*\;::program arguments: _normal'
+      '*-execdir:program: _command_names -e:*(\;|+)::program arguments: _normal'
```
danielshahaf, line 51: As below.

Completion/Unix/Command/_find
```diff
@@ -147,7 +150,7 @@ _arguments -C $args \
   '*-user:user:_users' \
   '*-xdev' \
   '*-a' '*-o' \
-  '(-D -E -H -L -O -P -f -s -x --help --version)*:directory:_files -/' \
+  '(-D -E -H -L -O -P -f -s -x -files0-from --help --version)*:directory:_files -/' \
```
danielshahaf, line 153: :+1:

danielshahaf:
> Note! I only have local access to GNU find. For the other platforms it is based on reading man pages. I have verified against man pages locally for GNU and man pages online for FreeBSD, OpenBSD and MacOS. I have not looked at other platforms as there are too many to go through, and I'm not sure if some of them are even still around.

That's fine; you're not expected to implement new functionality for platforms you don't have access to.  Thanks for documenting your research.

Completion/Unix/Command/_find
VorpalBlade: The best help text would require separate entries for each combination of flags.  Or code generating such separate entries.

Completion/Unix/Command/_find
VorpalBlade: So what is the right way to get it to work for + and ;?

VorpalBlade: I'm not sure what you mean is undocumented. Quoting the zsh documentation:
> This describes multiple arguments. Only the last optarg for an option taking multiple arguments may be given in this form. If the pattern is empty (i.e. :*:), all the remaining words on the line are to be completed as described by the action; otherwise, all the words up to and including a word matching the pattern are to be completed using the action.

Also, competing arguments for ls seem to work just fine for me?  I cannot replicate the issue you describe. That is using this file along with the latest release of zsh (5.8.1). I don't have zsh itself built from git. Maybe something else broke recently?

Completion/Unix/Command/_find
VorpalBlade: Again, this seems to be working perfectly for me, really confused as to what the issue could be.

Completion/Unix/Command/_find
danielshahaf: The potential problem with that is that it'd use up 16  screen lines, and vertical space isn't that cheap.  (Because the output is single column… it'd be nice to change that, if $COLUMNS is large enough, but that's a separate discussion.)

Completion/Unix/Command/_find
danielshahaf: I stand corrected: it's indeed documented as you say.

What I couldn't get to work is for `find . -exec ls <TAB>` to offer `\;` as a potential completion.  That's not a regression though so let's not worry about that.

danielshahaf:
I stand corrected: it's indeed documented as you say.

What I couldn't get to work is for `find . -exec ls <TAB>` to offer `\;` as a potential completion.  That's not a regression though so let's not worry about that.

I'm using recent master.

[ Sorry if you get this multiple times; github seems to have gone down between two of my comments. ]

Completion/Unix/Command/_find
danielshahaf: I stand corrected: it's indeed documented as you say.

What I couldn't get to work is for `find . -exec ls <TAB>` to offer `\;` as a potential completion.  That's not a regression though so let's not worry about that.

danielshahaf:
Reviewed again.  LGTM.  Anyone else has comments?

Completion/Unix/Command/_find
VorpalBlade: > What I couldn't get to work is for find . -exec ls <TAB> to offer \; as a potential completion. That's not a regression though so let's not worry about that.

Ah, now I see what you mean. I guess you could use (I haven't tested this!) `{_alternatives '_values \; +' _normal}` or something to that effect (but with valid syntax)

Completion/Unix/Command/_find
VorpalBlade: > [...] and vertical space isn't that cheap. (Because the output is single column… it'd be nice to change that, if $COLUMNS is large enough, but that's a separate discussion.)

I use [zsh4humans](https://github.com/romkatv/zsh4humans) which replaces the completion results with a tmux popup window with fzf search of the possible completions. I expect there are at least 10 other ways for users to completely replace how the completions are presented given that this is zsh.

Where am I going with this? Two points: 1) It is easy enough for users to change this if they want 2) Users *do* change things, so that limits what you can expect and assume anyway.

larryv: > Where am I going with this? Two points: 1) It is easy enough for users to change this if they want 2) Users _do_ change things, so that limits what you can expect and assume anyway.

No. Please do not justify bad default behavior by asserting that "users can just change it if they want".

VorpalBlade: > No. Please do not justify bad default behavior by asserting that "users can just change it if they want".

You are absolutely right, when you look at it from that point of view.

danielshahaf: IIRC, in this specific case the one-per-line behaviour is hardwired into `compdescribe`, the builtin [= implemented in C] guts of `_describe`, which `_arguments` calls.  (See uses of `zterm_columns` in [`bin_compdescribe()`](https://github.com/zsh-users/zsh/blob/master/Src/Zle/computil.c).)  This isn't something users can easily change, even if we assume `bin_compdescribe()` doesn't access zle's C internals in ways that a shell function can't.

zsh4humans and friends actually install a `compadd` wrapper function, which is problematic for reasons of composability (what if several plugins wanted to do that for different reasons?) and forward compatibility.

What the API _should_ be like is `zstyle :completion:settings:column-width 80` and then the screen would be divided into as many list-of-options columns as fit the terminal, each list-of-options column being 80 terminal columns wide.  Or perhaps `zstyle :completion:settings:preferred-completion-chooser myfunc` and then myfunc would get everything that's normally dumped to the terminal (group descriptions, candidate strings, candidate string descriptions; and all the forgoing for every element of tag-order) and would feed an answer back to compsys.

Completion/Unix/Command/_find
danielshahaf: Or perhaps with `_arguments` states and/or `_wanted terminator expl terminators compadd \; +`.

Anyway, as I said, it's not a regression.  If you get this to work, great; if not, it's not a blocker to merging the rest.

VorpalBlade: > ` _wanted terminator expl terminators compadd \; +.`

I have no idea what you just wrote. I clearly have a lot to learn still. ZSH argument completion is... dense. And so is the documentation. And most tutorials online only cover the basics. I really don't understand what _wanted is all about to be honest.

(Side note: I would love to see some more approachable documentation for the completion system as a whole. As it is now it is very dense reference documentation. Just adding examples of where you would want to use each feature (exists in some cases, but not all) would help a lot for example. Sure I can look at one of the existing completion files, but they are generally complex beasts, not made specifically to demonstrate a feature. Nor do they have the code comments that would be required in an illustrative example. Also `_arguments` and the other "big" ones would benefit from being restructured to have more sub-headings to split the documentation up so it is easier to find things in.)

okdana: > I would love to see some more approachable documentation for the completion system as a whole.

There are the [zsh-completions how-to](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org) and the [style guide](https://github.com/zsh-users/zsh/blob/master/Etc/completion-style-guide). Though even those aren't great imo. I had started writing my own guide at one point, but there's so much to cover and i tend to over-explain things so i worried that it'd just end up like the manual again...

(I haven't really looked at any of the changes or technical discussion here btw)

VorpalBlade:
I do not believe I will have time any time soon to look into the *pre-existing* issue of -exec not completing \; and +. So this is as good as it is going to get for now.

danielshahaf:
Pushed in a54d70457934662318feac9fcc750d2691250f33.  Missed 5.9, I'm afraid, but will be in the next release.

danielshahaf:
Thanks for the patch!

Completion/Unix/Command/_find
danielshahaf: @okdana I'd be happy to take a look at your draft, time permitting.  Feel free to send me a link (publicly or privately).

PR #92, opened 2022-05-17
fix: Use markdown for README

bevsxyz:
I have made some visual changes to make the readme more compatible with Github rendering.


okdana:
I'm not personally opposed to re-arranging these files, but it's a big change to a very long-standing convention, so i think it needs to be discussed on the mailing list. Could you post your proposal there (`zsh-workers at zsh.org`)?

larryv:
> I'm not personally opposed to re-arranging these files

I am. I don't see why we should markup-ify the readme for the benefit of viewing it on a platform that we don't even officially use.

bevsxyz:
I have added the following version badge that gets its version from the latest tag data provided by Github. This is in the JSON format, which is queried with https://shields.io/ to generate the badge. The badge version updation will be independent of the README.md source file.

[Version](https://github.com/zsh-users/zsh/tags)

README.md
```diff
@@ -1,4 +1,5 @@
-# THE Z SHELL (ZSH)
+
+# <img src="https://www.zsh.org/color_vertical_icon.png" alt="ZSH" width="50"/> THE Z SHELL (ZSH)
```
larryv, line 2: Our actual project host, SourceForge, [strips raw HTML when rendering Markdown](https://sourceforge.net/p/forge/documentation/Files-Readme/), so this won't work.

I remind you, again, that this project is not hosted on GitHub.

bevsxyz:
I have created a fork of zsh in Sourceforge just to see for myself how the readme looks. It looks fine.

https://sourceforge.net/p/zsh-fork/code/ci/master/tree/

README.md
bevsxyz: I have removed raw HTML and replaced it with a markdown image tag without styling.

FYI Sourceforge does not strip HTML. I may now have to revert the changes 😮‍💨 

https://sourceforge.net/p/zsh-fork/code/ci/a205af4809fbe2335438e6b83c1b55bdabba5504/tree/

larryv: > FYI Sourceforge does not strip HTML.

I guess their documentation is wrong, or I misunderstood it.

bevsxyz: That's okay. I will revert back to raw HTML only if we choose to go ahead with the PR. I hope that's alright.

PR #93, opened 2022-05-24
Replace /etc/mtab with /proc/self/mounts for Linux

nolange:
/proc/self/mounts has been available since Linux 2.4.19,
released in 2002. /etc/mtab is usually a symlink to this
file but might not exist.

Not sure how to deal with irix, should this just fallback to the default?


phy1729:
IRIX should be left as is as you did. Merged in c1f1bbc5b. Thanks!

phy1729:
IRIX should be left as is as you did. Merged in c1f1bbc5b. Thanks!

PR #94, opened 2022-05-24
support .zst kernel modules with modutils

nolange:


phy1729:
Merged in dd2c882ae. Thanks!

PR #95, opened 2022-08-13
Add completions for neomutt

Freed-Wu:
BTW, about <https://github.com/msys2/MSYS2-packages/issues/2965>, any method to
make zsh know cygwin's (or msys) start is not unix's start? Thanks.


Freed-Wu:
BTW, <https://github.com/pypa/pip/pull/11409> may remove <https://github.com/zsh-users/zsh/blob/master/Completion/Unix/Command/_pip>

PR #96, opened 2022-10-17
Fix a injection of 'args' variable.

nanorkyo:
SEE ALSO: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=267099

We found a storage behavior of 'cut' command's completion with https://github.com/wookayin/fzf-fasd . So I researched this issue and found injection of the 'args' variable defined in fzf-fasd. To fix this issue, I decided that 'args' variable is  renamed and typeset-ed.

### DIFF
--- a/Completion/Unix/Command/_cut
+++ b/Completion/Unix/Command/_cut
@@ -1,6 +1,7 @@
 #compdef cut gcut
 
 typeset -A _cut_args
+typeset -a _extra_args
 
 case $LANG in
   (de_DE.UTF-8)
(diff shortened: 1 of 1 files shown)


danielshahaf:
The freebsd issue mentions you found a "conflict of variable 'args'".  Where's the other variable $args that the one in `_cut` conflicted with?

danielshahaf:
Would you please link from the FreeBSD issue to this PR, to avoid duplication of triage work?  Thanks.

nanorkyo:
Thank you, I added that my PR link to FreeBSD issue.

okapia:
Rather than rename the variable, the correct fix is to declare args as local. I have commit that fix.

PR #97, opened 2023-01-05
Misc cleanup of all text files

basicfeatures:
- Remove CR+LF (^M)
- Remove trailing whitespace
- Compress two or more consecutive blank lines
- Ensure newline at EOF

https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file


larryv:
> this change would need to go to the mailing list

Before doing this, consider reviewing your proposals to see if they actually make sense.  For instance, blindly "[compressing] two or more consecutive blank lines" is undesirable because this code base frequently uses those for stylistic purposes.  You also propose modifying many files that should not be modified (e.g., changelogs).

If you do go to the list, I would suggest submitting several smaller patches instead of this elephantine megapatch -- one for each type of proposed change, perhaps.

PR #98, opened 2023-03-10
[Completion] add `shortcuts` completion

vidhanio:
Added zsh completion for the [`shortcuts` cli tool](https://support.apple.com/en-ca/guide/shortcuts-mac/apd455c82f02/mac) on Darwin.

### DIFF
--- a/Completion/Darwin/Command/_shortcuts
+++ b/Completion/Darwin/Command/_shortcuts
@@ -1,0 +1,87 @@
+#compdef shortcuts
+
+_shortcuts() {
+  local line state
+
+  _arguments -C \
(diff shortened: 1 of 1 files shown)


okapia:
This has been merged in 51d5ddb. Thanks for the contribution.

I have tweaked some descriptions to closer match zsh conventions: lowercase, no period and try to start with a verb.

There are areas where it could be improved, e.g. use `_call_program` which will work better on a system that doesn't have shortcuts available. Can options like `--output-path` be used with a following `=`? And I would guess that it is mutually exclusive with `-o` but this isn't marked.

PR #101, opened 2023-09-11
Add command to improve robustness

Freed-Wu:
Fix the following bug:

Screenshot from 2023-09-11 18-48-46

### DIFF
--- a/Completion/Unix/Type/_find_net_interfaces
+++ b/Completion/Unix/Type/_find_net_interfaces
@@ -12,7 +12,7 @@ PATH=/sbin:$PATH
 
 case $OSTYPE in
   aix*)
-    net_intf_list=( ${(f)"$(lsdev -C -c if -F 'name:description')"} )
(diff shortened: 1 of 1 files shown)


okapia:
It looks to me like you must have wrapper functions for `ip` and/or `ifconfig` to apply colours to the output. If you insist on doing this I would suggest either giving them different shorter and more specific names or perhaps altering the functions to only apply colouring if standard output corresponds to a terminal: `[[ -t 1 ]]` The onus is on whoever writes those wrappers to ensure robustness.

I'm not keen on very selectively adding `command` to just the one or two places where one user's setup is affected. Arguably we should be adding `command` everywhere but that gets a bit ridiculous. At least where, `_call_program` is used, you can set the `command` style to work around the problem.

I've posted to the mailing list (workers/52164). If you want to follow-up, the list will reach the relevant developers rather better than here.

Freed-Wu:
> The onus is on whoever writes those wrappers to ensure robustness.

I agree on your point.

PR #102, opened 2023-09-26
Ensure todo.sh completion use builtin commands

BuonOmo:
The `todo.sh` script allows redefining commands locally, which can mess up this completion script. We are using the `command` subcommand to avoid that.


phy1729:
Merged in 97e4310. Thanks!

PR #103, opened 2023-10-13
fix completion for zpool upgrade

christian-heusel:
There was a backslash missing for the completion of `zpool upgrade` which this PR adds!
Fixes the following error:
```
$ zpool upgrade _zfs:1239: command not found: (-v)-V+[upgrade to given version]:version
-v
```


phy1729:
Merged in 5fd8cccf5. Thanks!

PR #105, opened 2023-11-01
Add "-I" for "zle" completion

aoyama-val:


okapia:
Merged. Thank you for the contribution.

PR #106, opened 2023-11-12
update tmux completion (20230919); tweak Util/check-tmux-state

atomicstack:
- update tmux completion functions
- tweak check-tmux-state to take current file format into account


okapia:
Merged. Thank you for the contribution.

okapia:
However, I posted a follow-up patch to the mailing list: https://www.zsh.org/mla/workers/2023/msg01031.html
Would be good to have confirmation that that makes sense.

PR #107, opened 2023-12-01
cd: Append 2nd argument to CWD if 1st is empty

druckdev:
In cd's second form:

```
cd [ -qsLP ] old new
```

Currently when old is empty, cd will prepend new to the name of the current directory. As this has very limited use cases (e.g. `cd '' /proc` when in `/sys/...`), change the behaviour of cd to **append** new instead.

### DIFF
--- a/Doc/Zsh/builtins.yo
+++ b/Doc/Zsh/builtins.yo
@@ -287,7 +287,8 @@ directory hash table.
 
 The second form of tt(cd) substitutes the string var(new)
 for the string var(old) in the name of the current directory,
-and tries to change to this new directory.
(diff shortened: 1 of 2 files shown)


druckdev:
Since this could break some scripts relying on that behavior, the change could also be put behind an option that is off by default. But I first wanted to hear what the general opinion on this change is.

larryv:
Please submit a patch to zsh-workers@zsh.org for discussion, as described [here](https://zsh.sourceforge.io/Arc/git.html). This is an unofficial mirror which is not monitored by the primary developers.

druckdev:
Oh sorry, I did not read the part about only completion functions. I sent the patch to the mailing list.
Thank you!

PR #108, opened 2024-01-10
add perlbrew completion

orel1:
Hello,
Here you can find a completion script for [perlbrew](https://metacpan.org/pod/perlbrew).

### DIFF
--- a/Completion/Unix/Command/_perlbrew
+++ b/Completion/Unix/Command/_perlbrew
@@ -1,0 +1,316 @@
+#compdef perlbrew
+
+#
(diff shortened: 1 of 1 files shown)


okapia:
Thanks, with a few very minor tweaks, I've applied this in 0119a57e9.
I'm not familiar with perlbrew but the completion looks to be of good quality. Please consider revisiting it in future so that it doesn't become badly outdated over time.

The tweaks were to change a couple of group descriptions to singular form as per zsh convention - the thinking behind this is that description applies to what is completed rather than the list of matches as a whole.

Also removed the vim specific tags. I'm a vim user too but it not everyone is and we have a top-level `.editorconfig` file. There are vim plugins to honour those.

PR #109, opened 2024-01-11
Fix a typo

Freed-Wu:
$ valgrind --tool=massif --help | grep time-unit
    --time-unit=i|ms|B        time unit: instructions executed, milliseconds


phy1729:
Merged in dde1259. Thanks!

PR #110, opened 2024-01-18
Fixed incorrect completion for `dnf --showduplicates`

OpenSauce04:
The zsh completion for dnf's `--showduplicates` flag is incorrectly listed as `--show-duplicates`

image


okapia:
Applied in 742049a4
Thanks for the contribution

PR #111, opened 2024-02-02
Add new features and improvements to the `incarg` ZLE widget

midchildan:
This brings numerous improvements to the `incarg` ZLE widget that's used to increment an integer under the cursor. I find myself incrementing numbers very often when editing commands from shell history. I originally used `incarg` for this, but I missed a couple of features from Vim's increment command. The contents of this PR is what I've eventually settled on.

### Decrement integers without defining a new widget
Previously, the user had to define their own widget if they wanted to decrement instead of increment.

```zsh
decarg() {
  autoload -Uz incarg
  (( NUMERIC = -${NUMERIC:-${incarg:-1}} ))
  incarg
}
zle -N decarg
```

After this change, it's simply:

```zsh
zle -N decarg incarg
```

### Preserve the number of leading zeros
Previously, `incarg` omitted leading zeros.

```
Before: 001
After:  2
```

After this change, leading zeros are preserved.

```
Before: 001
After:  002
```

### Increment binaries, octals, and hexadecimals
Previously, `incarg` only recognized decimals. After this change, it'll recognize `0b`, `0o`, and `0x` prefixes. Both lower case and upper case prefixes are recognized.

### Move the cursor to the end of the incremented integer
Previously the cursor position didn't change after incrementing. This lead to annoyances when the number of digits increased:

```
Before: 99
          ^ cursor
After:  100
          ^ cursor
```

After this change, the cursor will be moved to the immediate right of the integer.

```
Before: 99
          ^ cursor
After:  100
           ^ cursor
```

This is probably a nice property even when the number of digits don't change. If I want to keep on editing the current line right after incrementing, I typically want to change something on the right of the integer.

### Create a sequence of integers across terminal panes
This adds a variation of the `incarg` widget that can be used by prefixing the widget name with `sync-`. It's intended to be used on multi-pane terminals with the keyboard input broadcast to all panes. It increments the integer in each pane by a different amount, with the next pane having its integer incremented once more than the previous one. Supported terminals are tmux and iTerm2.

```zsh
zle -N sync-incarg incarg
zle -N sync-decarg incarg
```

I often use this to interactively run the same commands on multiple sequentially numbered servers all at once. It's shown in the demo down below.

### Add a Vim variant
Vim's increment command jumps to the nearest integer after the cursor and increments it. I find this behavior very useful when combined with `history-beginning-search-backward`. So I made it possible to get this behavior by prefixing `vi` to the widget name.

```zsh
zle -N vim-incarg incarg
zle -N vim-decarg decarg
bindkey -a \
  '^A' vim-incarg \
  '^X' vim-decarg
```

It can also be combined with the `sync-` prefix.

```zsh
zle -N vim-sync-incarg incarg
zle -N vim-sync-decarg incarg
bindkey -a \
  'g^A' vim-sync-incarg \
  'g^X' vim-sync-decarg
```

## Demo
[asciicast](https://asciinema.org/a/ijvvGnXnqsgQn476JIplUxlp8)

### DIFF
--- a/Doc/Zsh/contrib.yo
+++ b/Doc/Zsh/contrib.yo
@@ -2620,12 +2620,30 @@ zle -N history-pattern-search-forward history-pattern-search)
 tindex(incarg)
 vindex(incarg, use of)
 item(tt(incarg))(
-Typing the keystrokes for this widget with the cursor placed on or to the
-left of an integer causes that integer to be incremented by one.  With a
-numeric argument, the number is incremented by the amount of the
-argument (decremented if the numeric argument is negative).  The shell
-parameter tt(incarg) may be set to change the default increment to
-something other than one.
+This widget allows you to increment integers on the current line. In addition
+to decimals, it can handle hexadecimals prefixed with tt(0x), binaries with
+tt(0b), and octals with tt(0o).
+
+By default, the target integer will be incremented by one. With a numberic
+argument, the integer is incremented by the amount of the argument. The shell
+parameter tt(incarg) may be set to change the default increment to something
+other than one.
+
+The behavior of this widget changes depending on the widget name.
+
+When the widget is named tt(incarg), the widget will increment an integer
+placed under the cursor placed or just to the left of it. tt(decarg), on the
+other hand, decrements the integer. When the name is prefixed with tt(vim-),
+the cursor will jump to the nearest integer after the cursor before incrementing
+it.
+
+There's also a tt(sync-) prefix that can be added to the widget name. This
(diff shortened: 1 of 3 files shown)


larryv:
Please submit a patch to zsh-workers@zsh.org for discussion, as described [here](https://zsh.sourceforge.io/Arc/git.html).  This is an unofficial mirror that is not monitored by the primary developers.

midchildan:
Sorry, I missed the unofficial status. I will resubmit this PR to the mailing list.

PR #114, opened 2024-02-29
remote files completion: remove double-escaping

fugidev:
Removes the double escaping in the remote files completion. This affects rsync and scp. Instead of for example a space character in a remote filename turning into `\\\ `, it will now correctly turn into `\ `. While scp apparently works with either one, rsync supports only the latter since version [3.2.4](https://download.samba.org/pub/rsync/NEWS#3.2.4). This has been a problem for almost two years now.

For more context, see the corresponding issue from fish-shell: https://github.com/fish-shell/fish-shell/issues/8895

### DIFF
--- a/Completion/Unix/Type/_remote_files
+++ b/Completion/Unix/Type/_remote_files
@@ -60,10 +60,7 @@ if zstyle -T ":completion:${curcontext}:files" remote-access; then
     dirprefix=${dir}/
   fi
 
-  if [[ -z $QIPREFIX ]]
-    then rempat="${dirprefix}${PREFIX%%[^./][^/]#}\*"
-    else rempat="${dirprefix}${(q)PREFIX%%[^./][^/]#}\*"
-  fi
+  rempat="${dirprefix}${(q)PREFIX%%[^./][^/]#}\*"
 
   # remote filenames
   remfiles=(${(M)${(f)"$(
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Do the zsh devs actively monitor Pull Requests here? Might be better on their mailing list?

(I'd love to see this merged, this bug is so annoying...)

larryv:
> Do the zsh devs actively monitor Pull Requests here?

This repository is an unofficial mirror.  While a few devs do check it on an irregular basis, the most active ones do not look at it at all.

> Might be better on their mailing list?

This is always the better option.

fugidev:
I have submitted the patch to the zsh-workers mailing list

ThiefMaster:
Any news on this?

larryv:
The discussion begins [here](https://www.zsh.org/mla/workers/2024/msg00446.html).  It seems to have stalled out.

ThiefMaster:
Ah, any chance it could be revived? I'd leave a comment on it myself but since it's a mailing list I'm not subscribed to, I wouldn't even know how to properly reply to that thread w/o having one of its emails... (yay for great usability in mailing lists!)

fugidev:
A (different) fix landed in https://github.com/zsh-users/zsh/commit/5cb32ec5ab1eafdad62500e7e6bdada2c3205a59 so I'm closing this.

PR #116, opened 2024-04-03
Fix copy-earlier-word handling of numeric arguments

huyz:
As mentioned in the comment:

```zsh
  # 1 means last word, 2 second last, etc.
```

the numeric argument is supposed to be interpreted as an offset from the end.

The intended behavior is a lot more useful than the current behavior


larryv:
Thanks for the contribution, but this is an unofficial mirror that is not monitored by the primary developers.  Please submit a patch to zsh-workers@zsh.org for discussion, as described [here](https://zsh.sourceforge.io/Arc/git.html).

huyz:
Gotcha

PR #117, opened 2024-04-22
completion(vim): vim also supports --clean option (since v8.2.2774)

e-kwsm:


okapia:
This patch moves `--clean` out of the block that is added for vim (but not nvim) and into the block that is unconditionally included regardless of `$service`. So is it actually nvim (neovim I assume) that has gained a `--clean` option? Have you got a reference for whatever changed in vim 9.1.0363? Otherwise, the change doesn't make sense to me for the description.

okapia:
Which version it was isn't really the point. The completion function already supported the `--clean` option in completion for vim. Your change only affects completion for neovim.

e-kwsm:
Ah, I see. I believe I addressed this on a machine that provides outdated completion file.
`vim --clean` completion works as expected.

PR #118, opened 2024-04-22
completion(git-rebase): non-interactive mode also supports --autosquash

e-kwsm:


okapia:
Merged in 8cddd97297

Sorry, it took me a while to get around to verifying this. I tweaked the change slightly because the extra variable is no longer needed. Thanks for being patient and for making the contribution.

PR #120, opened 2024-05-29
Update _git to reflect `--recursive` being an alias

Semnodime:
`--recursive` and `--recurse-submodules` are aliased according to https://github.com/git/git/blob/99c33bed562b41de6ce9bd3fd561303d39645048/builtin/clone.c#L105


okapia:
Merged and applied (with a slight adjustment). Thanks for the contribution.

PR #121, opened 2024-08-13
Completion: Add make variables with values

mstaz:
So far only variable names were added. However Makefiles and make would also return possible values which are not completed so far. The values are stored and overwritten in the map without using them anywhere. So instead it currently makes more sense to complete the whole variable name-value combination instead of only the name.

I'm unsure about this change because I have not completely understood zsh's completion system yet. However it works for me (version 5.8 and call-command set true) as I want it to be. It's similar to bash's behavior with that.

### DIFF
--- a/Completion/Unix/Command/_make
+++ b/Completion/Unix/Command/_make
@@ -76,7 +76,8 @@ _make-parseMakefile () {
     case "$input " in
       # VARIABLE = value OR VARIABLE ?= value
       ([[:alnum:]][[:alnum:]_]#[" "$TAB]#(\?|)=*)
-      var=${input%%[ $TAB]#(\?|)=*}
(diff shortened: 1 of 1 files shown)


larryv:
This contribution should probably be [discussed on the mailing list](https://zsh.sourceforge.io/Arc/git.html), since it significantly changes completion behavior. Very few committers monitor this repository mirror, and little to no discussion happens here.

mstaz:
Thanks for the advice. I followed [zsh-completions/CONTRIBUTING.md](https://github.com/zsh-users/zsh-completions/blob/master/CONTRIBUTING.md#contributing-completion-functions-to-zsh) which states that completion functions can be contributed via github pr as well. But maybe you are right.

PR #125, opened 2024-11-25
Add completion for irb --no-pager

kyanagi:
irb has `--no-pager` since https://github.com/ruby/irb/commit/df1c3b904284c4ec36924ca6deab2bcefc371eb1.


okapia:
Merged in 9f9ceb776

Thanks for the contribution.

PR #126, opened 2024-11-26
Fix some typos

Freed-Wu:


larryv:
Thanks for the suggestions. In general, only contributions to completion files are accepted via PR on this mirror, which is unofficial. Please email a patch to zsh-workers@zsh.org instead.

okapia:
Despite not being to the completions, this shouldn't be too controversial so I merged in a61b105795.

Thanks

okapia:
For future reference, I should mention that it would have been better if the title/commit message of the PR mentioned that these were typos **in comments**

PR #128, opened 2025-01-21
_vim: Updated completion for neovim

gi1242:
Added various `--remote` options that neovim supports, and correctly complete neovim servers. (Neovim uses `--server` instead of `--servername`)


okapia:
Merged in 45b79fa2cb

Thanks

PR #129, opened 2025-02-19
fix _gpg completion

msva:
1) THe completion of `--local-user` wass irrelevant: it completed system users, while gpg expects private keys IDs

2) `secret-keys` completion was also broken and completed nothing. The reason of that to happen was in that fact that it assumed `uid` token would be next after `fpr` (same as it do in public-keys completion), while in current gnupg versions there is at least `grp` token, and potentially can be others. So, instead i+=2 I made `until "uid"` loop.

### DIFF
--- a/Completion/Unix/Command/_gpg
+++ b/Completion/Unix/Command/_gpg
@@ -29,7 +29,7 @@ fi
   '(-c --symmetric)'{-c,--symmetric}'[encrypt with symmetric cipher only]'
   '(-s --sign)'{-s,--sign}'[make a signature]'
   '*'{-r+,--recipient}'[specify user to encrypt for]:recipient:->public-keys'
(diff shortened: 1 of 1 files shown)


msva:
(I've heard that in is possible to contribute completions (and fixes for them) thru github, and it it not necessary to post patches to mail list for that)

larryv:
That's technically true, but in practice only a couple of contributors check for PRs here, and they do so irregularly.

okapia:
Merged in 1818323f4
Thanks for the contribution.

Using github is fine for completions. I pushed for that exception on the basis that completions tend to be fairly independent, don't need discussion on the mailing list and contributions often come from one-off contributors. I tend to check for PRs whenever I have something else to push and sometimes someone else gets there first. But as Larry mentions, this can be somewhat irregular.

PR #131, opened 2025-04-12
completions

crpb:
My Maintainer is to lazy so here goes..
> https://salsa.debian.org/debian/zsh/-/merge_requests/8#note_604077


crpb:
hey @okdana ,

on the use of `eval` i have to laugh as i used that first and then @ft said i should not do it :p.
anyhow, thanks for taking the time and improving it as i'm more guessing what i do their than anything else but wanted most of those changes finally be not only in my $fpath :p.

crpb:
> sorry, it did occur to me that it might've been a deliberate choice. if there were objections to it i can put it back

leave it, there are more of them in the completion :P.



> isn't it too late for 5.10 to go into trixie?

no idea, but he should be able to say as the packager

mika:
> > on the use of `eval` i have to laugh as i used that first and then @\ft said i should not do it :p
> 
> sorry, it did occur to me that it might've been a deliberate choice. if there were objections to it i can put it back
> 
> > it would benefit the soon to be debian stable zsh release :-). @\mika 👋🏼
> 
> isn't it too late for 5.10 to go into trixie?

We're currently in the soft freeze stage (see https://release.debian.org/testing/freeze_policy.html#soft) and 11 days before hard-freeze (2025-05-15). I don't know how large and instrusive the changes between 5.9 and 5.10 will actually be, but given that last stable release v5.9 dates back 2022-05-14 (AKA ~3 years), AFAICS that will be quite something :tm: overall. I'd assume zsh 5.10 also wouldn't show up within the very next few days, nor? :thinking: 

So from my current PoV we probably won't see zsh 5.10 with trixie. :-/

fivemeyestore:
Interesting solution to this problem! I've been working on similar challenges recently.

crpb:
@okdana 
today i realized that the completion with e.g. `apt-get source` isn't implemented at all for only source packages. i tinkered a bit with it but would kindly ask if you could add that aswell because the logic of retrieving those was alreadty [added](https://github.com/zsh-users/zsh/commit/98f670cf43b5ab4ee584487a385409c7c4f0c8c4) 9 years ago but never used in `_apt` as it seems.
What i'm talking about is `compdef '_deb_packages source' moo`. 
So the commands like `apt[-get] (source|build-dep) apt-se<TAB>` should complete **apt-setup** which is a source-package from the installer and not usable in any other way(like install/..) afaik.

I got the completion working with `_apt-cmd` but in that `_apt-get` i'm again struggling so i thought this would be easier this way 🙈 🙊 🙉

crpb:
> idk what the clean/idiomatic way to do that with `_regex_arguments` is though. tbh i can't stand it and i don't want to learn. rather than just copy+paste everything i'll have to let someone more knowledgeable figure out how to structure/format it

hah! i struggle with that extremly as i want to extend the [wpctl completion](https://gitlab.freedesktop.org/crpb/wireplumber/-/compare/master...master?from_project_id=2941). not my latest local thing but you will see silly it looks :P

If you know someone please ask them.

okdana:
thanks for taking care of this. sorry i missed the earlier reply, and sorry i made the pr more complicated than it needed to be. i really expected 5.10 to be finishing up at the time

PR #134, opened 2025-05-17
__git_worktrees: Use a library function

danielshahaf:
Using _describe makes the completions and descriptions line up in columns, and makes the function honour the list-separator style.

### DIFF
--- a/Completion/Unix/Command/_git
+++ b/Completion/Unix/Command/_git
@@ -8527,9 +8527,10 @@ __git_worktrees () {
       branch="[$branch]"
     fi
 
-    descriptions+=( "${directories[-1]}"$'\t'"$hash $branch" )
+    descriptions+=( "${directories[-1]//(#b)([\\:])/\\${match[1]}}":"$hash $branch" )
   done
-  _wanted directories expl 'working tree' compadd -ld descriptions -S ' ' -f -M 'r:|/=* r:|=*' -a directories
+
+  _describe -t directories 'working tree' descriptions -S ' ' -f -M 'r:|/=* r:|=*'
 }
 
 (( $+functions[__git_difftools] )) ||


danielshahaf:
> should you `local -a match mbegin mend`?
> 

Yes, I should, in case a caller does a pattern match and calls `__git_worktrees` before examining their backreferences.

Fixed.

> otherwise lgtm
> 

Thanks; I'll push (with the fix) shortly.

> ps: in workers/53577 i changed it (or i will do, i'm trying not to mess with things too much before release) so that all of the places i knew of where it completed or described revs did so using the format of `__git_recent_commits`, which is like `[$hash] $description`. i guess it's not perfectly analogous but i wonder if this would be another candidate for that

What change are you proposing—to put square brackets around the sha1, or to show the commit subject (= the commit's first line, `git log --pretty=%s`) rather than [the worktree HEAD's symbolic referent or the string "detached HEAD"]?

Addressing the two possible interpretations in logical order:

### What to show: HEAD commit's subject? HEAD's symbolic referent (if any)?

>> Should `__git_worktrees` show HEAD's symbolic referent?

As you may know, it's possible to have any number of worktrees of `master^{commit}`, but only one worktree of `master`.

Therefore, I'd say `__git_worktrees`' output should clearly indicate when a worktree's HEAD is a symbolic ref.

>> Does indicating that a worktree's HEAD is a symbolic ref require printing the referent?

I imagine that if we don't print the referent we'll instead print HEAD's sha1 or `git name-rev` output, which would amount to basically the same thing; thus, we might as well print the referent.

>> Should `__git_worktrees` indicate the commit's subject _in addition_ to what it already prints?

I think this would be useful in some cases, but less so in others:

- If the worktree is of a release branch, it's likely more useful to say "This is a worktree for the _foo_ branch" than to show the commit that's the current tip of the _foo_ branch. The latest commit on a release branch might be a random bugfix not be worth spending terminal real estate [cells in the 80x24] on in this context.

- If the worktree is a local branch actively worked on, showing the latest commit's subject could easily be useful.

Consequently, I think this might fall under ["In the face of ambiguity, refuse the temptation to guess"](https://peps.python.org/pep-0020/#the-zen-of-python).  This would point towards having a predictable, simple-to-explain default behaviour, and allow customization if need be.  For instance, we could have `__git_worktrees`' default output be whatever `git worktree list` does.

\* *danielsh runs `git worktree list` in one terminal and invokes `__git_worktrees` in another terminal*

Well, it turns out that "`__git_worktrees`' default output" already is "whatever `git worktree list` does"—up to the order of the lines.

While discussion on the candidate description strings continues, I'll quickly add a `-V` into `__git_worktrees`, so the order of candidates would match `git worktree list`'s output.  I hope that won't break anybody's proverbial spacebar heating…

>> Is there anything else the worktree description could include?

When describing a worktree—whether in completion, in `vcs_info`, or elsewhere—there's a number of data that the user might find useful:

- [discussed above] The worktree HEAD's symbolic referent.

- [discussed above] The worktree HEAD commit's subject.

- A gitrevisions(7) expression pointing to the worktree's HEAD.

    - [if the worktree is partway through an interactive rebase] A gitrevisions(7) expression pointing to the commit currently being `pick`-ed, or to the value of HEAD before the rebase started.

- Cribbing from [the list of `vcs_info` expandos](https://github.com/zsh-users/zsh/blob/fceece6a8a6a006b1ba546882aea10f9e03b0970/Doc/Zsh/contrib.yo#L1277-L1326):

    - %a: whether running git-rebase(1), git-merge(1), or git-cherry-pick(1) with `--abort` would make sense.  (These situations—unresolved conflicts, interactive rebases, and so on—are generally indicated by `git status` without `-s`.)

    - %c/%u: whether there are any staged changes or any changes at all.

    - %m/%Q: information about the patch series (quilt, stg, git-rebase(1), etc.) currently checked out.

### Whether to show brackets around the commit hash

If we show the hash and subject, we should show them as we do elsewhere, for consistency.

However, so long as we show the hash and HEAD symbolic referent, I see two arguments against adding brackets: one, it would break consistency with `git worktree list`'s output for little gain; two, our own output in different cases should be different enough that it's easy to tell at a glance which case is before one; thus, in particular, a hash that's _not_ followed by a commit subject should look different to one that is.

### Bottom line

I hope that answers your question :-)

Thanks again for the review!

danielshahaf:
Merged in 2ccf41e57.  `-V` added in f24958a7a.

danielshahaf:
(We can continue the discussion about commit subjects, of course; I just closed the PR since it has been merged, and I'd like to keep the open PRs list uncluttered.)

danielshahaf:
> if we *did* add the `HEAD` commit subject, which might be useful, then 
> it would be closer. maybe that could look like `[123456789] (master) 
> commit subject`, if there's not an existing format for that

It sounds like `git log --oneline --decorate`.  Looking it up, I see there's also `git log --pretty=reference`.  These formats indicate the hash, subject, refs, and (the latter format only) the author date.

PR #135, opened 2025-05-19
53605: Add tips in `compadd` docs about groups

RocketMaDev:
I've sent an email to mail list, but no one reply to me,
I got no idea if there is something wrong with my mail or what,
so I have to post my solution here to seek reply.

Main thread: https://www.zsh.org/mla/workers/2025/msg00307.html (zsh-workers 53605)


larryv:
There was nothing wrong with your email; no one's replied to it yet, is all.

PR #140, opened 2025-06-29
completion: remove `-a` from doas that is no longer supported

e-kwsm:


phy1729:
`-a` is still supported on [OpenBSD](https://man.openbsd.org/doas).

e-kwsm:
A forked doas do not support the option, so I thought it is removed.
Thank you for pointing it out.

danielshahaf:
It may be possible to support both with _pick_variant; see, for instance, _watch.

PR #145, opened 2025-09-24
update completion: _install

rrotter:
current to:
DragonFly 6.4.2
FreeBSD	14.3
NetBSD 10.1
OpenBSD 7.7
macOS 15.6.1

- -S[temp file] removed except for DragonFly; all other BSDs have removed this flag, and I can't find a commercial Unix that supports it
- -S[flush] replaces old -S opt on Darwin, FreeBSD
- -S disabled on OpenBSD which still supports it as a noop/compat option

- corrected +N flag to -N

- remove -C for NetBSD, it appears that this was never supported

- several other updates (mostly Net/Free flags added to Darwin/DragonFly)


okapia:
Merged in 32d1158b

Thanks

PR #162, opened 2026-03-09
completion: shutdown: suggest 'now' for time argument

nickshiro:
Current shutdown completion uses `_guard` for the time argument:

    '1: :_guard "^-*" "time (now/hh:mm/+mins)"'

`_guard` only validates the argument but does not provide completion
candidates. As a result, usage such as:

    shutdown n<TAB>

does not complete to `shutdown now`.

This patch introduces a completion state for the time argument and
adds a suggestion for the commonly used value `now`. The completion
still allows arbitrary values such as `+mins` and `hh:mm`.

### DIFF
--- a/Completion/Unix/Command/_shutdown
+++ b/Completion/Unix/Command/_shutdown
@@ -1,9 +1,10 @@
 #compdef shutdown
 
 local -a args
+local state
 
 if [[ -d /etc/systemd ]]; then
-  _arguments \
+  _arguments -C \
     '--help[display usage information]' \
     '(-H --halt)'{-H,--halt}'[halt the machine]' \
     '(-P --poweroff -h)'{-P,--poweroff,-h}'[power-off the machine (default)]' \
@@ -11,8 +12,19 @@ if [[ -d /etc/systemd ]]; then
     "(--no-wall)-k[don't shutdown, just write wall message]" \
     "(-k)--no-wall[don't send a wall message]" \
     '-c[cancel pending shutdown]' \
-    '1: :_guard "^-*" "time (now/hh\:mm/+mins)"' \
+    '1:time:->time' \
     '*:warning message'
+
+  case $state in
+    time)
+      _values -s ' ' 'shutdown time' \
+        'now[shutdown immediately]' \
+        '+1[shutdown in 1 minute]' \
+        '+5[shutdown in 5 minutes]' \
+        '+10[shutdown in 10 minutes]'
+      ;;
+  esac
+
   return
 fi
 
@@ -23,7 +35,7 @@ case $OSTYPE in
       '-r[reboot the system]'
       '-k[kick everybody off]'
       '-n[prevent file system cache from being flushed]'
-      '1: :_guard "^-*" "time (now/hh\:mm/+mins)"'
+      '1:time:->time'
       '*:warning message'
     )
   ;|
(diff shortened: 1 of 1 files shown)


arkriny (commented):
Isn't the same behavior achievable with a simpler change like this?

```diff
diff --git a/Completion/Unix/Command/_shutdown b/Completion/Unix/Command/_shutdown
index a237b14e0..a31990a4a 100644
--- a/Completion/Unix/Command/_shutdown
+++ b/Completion/Unix/Command/_shutdown
@@ -11,7 +11,7 @@ if [[ -d /etc/systemd ]]; then
     "(--no-wall)-k[don't shutdown, just write wall message]" \
     "(-k)--no-wall[don't send a wall message]" \
     '-c[cancel pending shutdown]' \
-    '1: :_guard "^-*" "time (now/hh\:mm/+mins)"' \
+    '1: :{ _guard "^-*" "time (now/hh:mm/+mins)" && compadd now }' \
     '*:warning message'
   return
 fi
@@ -23,7 +23,7 @@ case $OSTYPE in
       '-r[reboot the system]'
       '-k[kick everybody off]'
       '-n[prevent file system cache from being flushed]'
-      '1: :_guard "^-*" "time (now/hh\:mm/+mins)"'
+      '1: :{ _guard "^-*" "time (now/hh:mm/+mins)" && compadd now }' \
       '*:warning message'
     )
   ;|
```

This preserves _guard behavior while also adding completion for 'now'.

While I haven't found similar uses of compadd inline in _arguments, that's probably because it's a relatively rare case where an argument takes on both known and arbitrary values.

okdana:
> Isn't the same behavior achievable with a simpler change like this?

this only works when there's a partial match, as in `shutdown n<TAB>`. it won't offer it for `shutdown <TAB>` like the original change. i guess it's an improvement but doesn't seem ideal, especially since `now` is probably what you want 90% of the time

idk about the original change either. _values is the wrong thing to use for this, it crosses the line for copy/paste imo, and hard-coding specific examples of arbitrary times feels a little weird to me

this is probably the closest to what other functions do in cases like this, though it has the opposite problem of not telling you you can give a time in some cases

```diff
diff --git a/Completion/Unix/Command/_shutdown b/Completion/Unix/Command/_shutdown
index a237b14e0..946dbd440 100644
--- a/Completion/Unix/Command/_shutdown
+++ b/Completion/Unix/Command/_shutdown
@@ -1,5 +1,18 @@
 #compdef shutdown
 
+# note: this function is most accurate to the FreeBSD/Darwin implementation,
+# which is generally a super-set of the others
+(( $+functions[_shutdown_times] )) ||
+_shutdown_times() {
+  if compset -P +; then
+    _numbers -t times-plus -u minutes time s:seconds :m:minutes h:hours
+  else
+    _alternative \
+      'times-now:time (now):(now)' \
+      'times-expr: :_guard "^n*" "time (yyyymmddhhmm/hh:mm/+mins)"'
+  fi
+}
+
 local -a args
 
 if [[ -d /etc/systemd ]]; then
@@ -11,7 +24,7 @@ if [[ -d /etc/systemd ]]; then
     "(--no-wall)-k[don't shutdown, just write wall message]" \
     "(-k)--no-wall[don't send a wall message]" \
     '-c[cancel pending shutdown]' \
-    '1: :_guard "^-*" "time (now/hh\:mm/+mins)"' \
+    '1: :_shutdown_times' \
     '*:warning message'
   return
 fi
@@ -23,7 +36,7 @@ case $OSTYPE in
       '-r[reboot the system]'
       '-k[kick everybody off]'
       '-n[prevent file system cache from being flushed]'
-      '1: :_guard "^-*" "time (now/hh\:mm/+mins)"'
+      '1: :_shutdown_times' \
       '*:warning message'
     )
   ;|
```

okdana:
oliver addressed this in a different way with b0e2c191ae33

PR #165, opened 2026-03-24
New completion for ImageMagick 7 (magick command)

robbyrob42:
ImageMagick 7 unified all tools under a single `magick` command, replacing the standalone binaries from IM6 (convert, identify, mogrify, etc.). The existing `_imagemagick` completion covers IM6 but has no support for the `magick` command or IM7's option set.

This adds `_magick` which:
- Completes `magick` and all legacy tool names via `#compdef`
- Dispatches `magick <tool>` subcommands (identify, compare, etc.) to tool-specific option sets
- Handles bare `magick [options...] input output` (convert-style) usage
- Completes enum values for colorspace, compose, compress, filter, gravity, metric, morphology, distort, evaluate, layers, and more — all sourced from `magick -help` and `magick -list` output
- Filters file completion to common image formats with fallback to all files
- Uses `_pick_variant` to detect IM7 and falls back to `_files` on IM6, so it coexists safely with `_imagemagick`
- Supports `_call_function` hooks for user-overridable subcommand completions

Tested against ImageMagick 7.1.2-17 on macOS (Homebrew). Follows the completion style guide conventions.

Note: `_imagemagick` and `_magick` both register `#compdef` for the legacy binary names (convert, identify, etc.). On systems with IM7, the zsh maintainers may want to decide how to handle the overlap — e.g., updating `_imagemagick` to defer to `_magick` when IM7 is detected, or deprecating `_imagemagick` in favor of this file.

### DIFF
--- a/Completion/Unix/Command/_magick
+++ b/Completion/Unix/Command/_magick
@@ -1,0 +1,612 @@
+#compdef magick magick-script convert mogrify identify composite compare montage stream display animate import conjure
+
+# Completion for ImageMagick 7.1 (https://imagemagick.org).
+# ImageMagick 7 unifies all tools under the magick command.
+
+# Require ImageMagick 7+. Fall back to default file completion for IM6 or
+# older so we don't offer IM7-specific options to the wrong version.
+# For legacy binary names (convert, identify, etc.) we probe `magick` if
+# available, otherwise the service command itself.
+local _im_variant _im_pick_cmd
+if (( $+commands[magick] )); then
+  _im_pick_cmd=magick
+else
+  _im_pick_cmd=$service
+fi
+_pick_variant -c $_im_pick_cmd -r _im_variant \
+    im7='ImageMagick 7' im6='ImageMagick' unknown -version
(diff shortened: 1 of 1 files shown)


okapia:
Our naming convention is that a completion for, e.g. imagemagick is called `_imagemagick`, regardless of what the command is called. So e.g. subversion is `_subversion`, not `_svn` and `_libvirt` is the function for `virsh`. Mostly, follow the naming used for Linux etc packages. Also on naming, if separate functions are used for subcommands, use a hyphen, e.g. _magick-convert. Use underscores for helpers,, e.g. _imagemagick_files. However, I don't see anything wrong with the approach taken by the existing _imagemagick where it isn't broken into smaller functions.

It'd be easier to review this, if it was based on the old function because we now need to check whether any old functionality that would still be applicable is being lost. I wouldn't bother checking for ImageMagick 6 at the top, Just complete 7 options and if there are a few extra then so-be-it. That's more useful than degrading to just file completion.

robbyrob42:
appreciate the feedback, i'll iterate and come back

PR #166, opened 2026-03-25
Autocomplete FreeBSD Jail Names

DominoTree:
This is a bit convoluted because FreeBSD does not track a ton of jail state in the kernel, so for non-running jails we have to pull their names from the system's jail config file(s) - `sysrc -n jail_conf` should be the canonical way to do this


Completion/BSD/Command/_init_d-jail
```diff
@@ -0,0 +1,18 @@
+#autoload
+
+# Complete jail names for "service jail <action> [jail ...]"
+# Called by _init_d when the service script is "jail".
+# $1 is the action (start, stop, restart, etc.)
+
+local action=$1
```
okdana: superfluous

okdana (approved):
this looks good but i'll give it a few more days in case anyone has an opinion on my feedback

okapia:
When trying this, I get lots of errors of the form:
```
_jails:25: failed to compile regex: trailing backslash (\)
```
Zsh's own pattern matching is powerful enough and less likely to hit platform specific differences so I'd recommend just using that instead of `=~`. We virtually never use `=~` in the completion system, the few examples not having come from core developers. The backreference match variables should be declared as local.

It'd be good to complete configured jails also after `jail -c`. (But not jail -rc)

How common is extra service specific arguments? I've only seen it on FreeBSD and mostly used with `service netif restart`. if there's only a couple of uses, @okdana may be right that it doesn't need extra functions.

DominoTree:
Thanks both for taking a look here - this is my first time taking a swing at doing any sort of ZSH autocompletion

I've seen this pattern in a few places, but working with jails seems by far to be the most useful place to add autocompletion that I've observed

I'll see if I can incorporate this feedback and I'll also take a look to see how common this pattern is across the other default rc scripts

DominoTree:
> How common is extra service specific arguments? I've only seen it on FreeBSD and mostly used with `service netif restart`. if there's only a couple of uses, @okdana may be right that it doesn't need extra functions.

At a glance it looks like these rc scripts in the base system accept extra arguments - no idea how common this pattern might be in third-party packages but I don't recall ever noticing it

```
/etc/rc.d/ipfw_netflow
/etc/rc.d/rfcomm_pppd_server
/etc/rc.d/jail
/etc/rc.d/netif
/etc/rc.d/ipfw
/etc/rc.d/dnctl
/etc/rc.d/ppp
/etc/rc.d/syscons
/etc/rc.d/moused
/etc/rc.d/virtual_oss
/etc/rc.d/bluetooth
/etc/rc.d/routing
/etc/rc.d/msconvd
```

DominoTree:
@okapia could you take a look again when you get a minute?  I cleaned things up and used glob patterns and added support for `jail(8)`

Completion/Unix/Command/_init_d
```diff
@@ -122,4 +122,14 @@ cmds=( $(_init_d_get_cmds) ) || return 1
 (( $#cmds )) || zstyle -a ":completion:${curcontext}:commands" commands cmds ||
     cmds=(start stop)
 
-_arguments -s -A "-*" $flags ':init.d command:_sub_commands $cmds'
+local svcname=$words[1]
```
DominoTree: Unnecessary but seemed cleaner

Completion/BSD/Command/_init_d-jail
```diff
@@ -0,0 +1,18 @@
+#autoload
```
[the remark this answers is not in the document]
DominoTree: looks like there are ~14 separate services on a bare FreeBSD installation that support this syntax

okapia:
There's no need to do `setopt localoptions extendedglob`. For completion functions, the options in `$_comp_options` are set.

Running jails should also be completed after `-R`. That's an existing problem rather than something you introduced.

I still basically agree with the original point by @okdana that the separate definitions for extra arguments is too trivial to warrant splitting out. Rather than testing `(( $+functions[_init_d-$svcname] ))` there's a helper function named `_call_function`. For the trivial case of jail (and netif), a case-statement that calls _jails / _net_interfaces directly would suffice. It could use `_call_function` otherwise if you think it is useful to allow users to plug-in an extra function. But it is fairly obscure and FreeBSD specific as far as I can tell. None of the service files on OpenBSD seem to support extra arguments.

Otherwise, looks good. It parses my jail.conf fine but that may not be the most complex example out there.

llua:
```
name
{
...
}
```
is an syntactically correct jail definition that the pattern matching won't match.

okapia:
Just a brief ping to ask whether you were likely to push an update addressing the later review points.

llua's point is definitely valid but don't feel you have to handle every possible contortion within `jail.conf`. Silently failing to add a jail as a match is still acceptable. But if it's easy to do, there's no harm.

We're currently preparing things for a new minor release (including completion updates) and it would be nice to include this. First test release might be readied within days.

DominoTree:
> We're currently preparing things for a new minor release (including completion updates) and it would be nice to include this. First test release might be readied within days.

Been a bit swamped but I will get this sorted in the next ~day, would love to get it included!

DominoTree:
I think this should do it

okapia:
Merged in 2e6560e4.

Thanks for the contribution!

PR #168, opened 2026-04-30
_apt add history commands

crpb:
Extend apt completion with new commands (apt>=3.1.6): history-list, history-info, history-redo, history-undo and history-rollback


okapia:
Merged. Thanks!

PR #169, opened 2026-05-04
fix darcs completion script

glk0:
`darcs <something> --list-options`  provides some of the available options in the format <option>;<docs> . the former script dumps the full line, which is quite incovenient and you have to delte the pasted doc if you select an option. This fix replace the ';' with a ':' to display the docs properly and select only the option.

### DIFF
--- a/Completion/Unix/Command/_darcs
+++ b/Completion/Unix/Command/_darcs
@@ -31,8 +31,9 @@ else
         ;;
     # Otherwise, let's ask darcs for all possible options
     *)
-      _wanted args expl 'arg for darcs command' \
-        compadd -- $( darcs ${words[2]} --list-option )
+        local -a darcs_opts
+        darcs_opts=( ${(@)${(f)"$(darcs ${words[2]} --list-options)"}//;/:} )
+        _describe -t args 'arg for darcs command' darcs_opts
       ;;
   esac
 fi


Completion/Unix/Command/_darcs
```diff
@@ -31,8 +31,9 @@ else
         ;;
     # Otherwise, let's ask darcs for all possible options
     *)
-      _wanted args expl 'arg for darcs command' \
-        compadd -- $( darcs ${words[2]} --list-option )
+        local -a darcs_opts
```
okdana, line 34: this might be a little better

```
local MATCH MBEGIN MEND
local -a darcs_opts
darcs_opts=( ${(f)"$(
  _call_program ${(Q)words[2]}-options $words[1] $words[2] --list-options
)"} )
darcs_opts=( ${${darcs_opts//(#m)[][:]/\\$MATCH}/(#m);*/\[${MATCH#?}\]} )
(( CURRENT-- ))
shift words
_arguments : $darcs_opts
```

glk0: actually, in the meantime I found that Darcs provided its own Zsh completion script at https://hub.darcs.net/darcs/darcs-screened/browse/contrib/_darcs.zsh that solved the same issue a while ago. I think it would be better to integrate that one instead.

okdana: with only a few exceptions, when a project ships their own function we tend to just let them handle it and remove ours entirely

it looks like arch and pkgsrc correctly package the up-stream function, but debian and homebrew don't. fedora includes the file but it's in the wrong place

idk

glk0: I think we should go with copying the darcs-maintained script  here. The 
darcs project doesn't receive a lot of maintainance (not so popular nowadays), the completion script hasn't been touched in years, so just keeping that copy improves the UX for the (few) Darcs users there are, and there won't be problems with keeping the completion in script in sync with the darcs one (at leaste for quite a while). Should I push that script in my PR ?

okdana: @okapia wdyt?

okapia: Usually, we drop our completion if upstream provide one that is either the same function or a better one.

In this case, the darcs completion appears to be covered by the GPL which provides an extra reason to drop it. We should avoid having completions covered by different licences, not because of rabid anti-GPL opinions or such reasons but because creating a complicated licencing situation creates bureaucratic hassles for many distributors and packagers - some of whom do even remove individual functions. Is less work overall to rewrite the functions.

If you want, update it in the stable (5.9.1) branch and lose it from master.

okdana:
updated from up-stream (99d875d6) then removed (1502edd5) in master

cherry-picked update into 5.9.1 (b9e7be81)

if your os/distribution doesn't package the up-stream function properly feel free to report it to them, should be an easy fix

PR #170, opened 2026-05-11
Clarify the documentation about -d flag in compadd

pacak:
I spent a lot of time trying to understand why won't my completions work as I would expect them to. Hopefully this clarifies things a bit.

### DIFF
--- a/Doc/Zsh/compwid.yo
+++ b/Doc/Zsh/compwid.yo
@@ -520,8 +520,12 @@ item(tt(-d) var(array))(
 This adds per-completion display strings. The var(array) should contain one
 element per var(completion) given. The completion code will then display the
 first element instead of the first var(completion), and so on. The
-var(array) may be given as the name of an array parameter or directly
-as a space-separated list of words in parentheses.
+var(array) may be given as the name of an array parameter, or directly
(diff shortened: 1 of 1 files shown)


pacak:
Parser in question lives here

https://github.com/zsh-users/zsh/blob/ef574dbb79a4a0bcdaa02819954b39c9614b4b05/Src/Zle/compcore.c#L1971-L1983

okdana:
i didn't realise it used that syntax either, i assumed it was just the rhs of an `arr=(...)` assignment like with _arguments, _describe, etc

looks ok to me at first glance but i'll give others time to comment before i merge

pacak:
Syntax that works is this one, gives several lines with --alpha, --beta, etc along with the descriptions. Feel free to play around.

```
compadd -l -d "(--alpha\ --\ Alpha:\ α --beta\ --\ Beta:\ β --gamma\ --\ Gamma:\ γ -z\ --\ Zetta:\ ζ)" -- --alpha --beta --gamma -z
```

pacak:
As far as I understand `compadd` is a builtin thing so it does its own parsing. `_describe` and others use shell to do the escaping.

okdana:
merged as 28c3808d. i adjusted the quote placement slightly

PR #172, opened 2026-05-13
Completion for "sioyek" (PDF viewer)

gi1242:
Sioyek is a PDF viewer with a focus on technical books and research papers.

https://sioyek.info/

This script provides a completion file

### DIFF
--- a/Completion/X/Command/_sioyek
+++ b/Completion/X/Command/_sioyek
@@ -1,0 +1,29 @@
+#compdef sioyek
+
+_arguments -S \
+  '--new-instance[create a new instance of sioyek when opening a file]' \
+  '--instance-name[select a specific sioyek instance by name]:name: ' \
+  '--new-window[open file in a new window within same instance]' \
+  '--reuse-window[force reuse of current window]' \
+  '--nofocus[do not bring instance to foreground]' \
+  '--version[print sioyek version]' \
+  '--page[which page to open]:page number: ' \
+  '--focus-text[set a visual mark on line including text]:text: ' \
+  '--focus-text-page[page to use for focus-text]:page number: ' \
+  '--inverse-search[command to execute for inverse search]:command: ' \
+  '--execute-command[command to execute on running instance]:command: ' \
+  '--execute-command-data[optional data for execute-command]:data: ' \
+  '--forward-search-file[forward search on file]:file:_files' \
+  '--forward-search-line[forward search on line]:line: ' \
+  '--forward-search-column[forward search on column]:column: ' \
+  '--zoom[set zoom level]:zoom level: ' \
+  '--xloc[set x position within page]:x-coordinate: ' \
+  '--yloc[set y position within page]:y-coordinate: ' \
+  '--window-id[apply command to specific window id]:window id: ' \
+  '--shared-database-path[path for shared bookmarks/highlights]:database path:_files' \
+  '--verbose[print extra information]' \
+  '--wait-for-response[wait for command to finish before returning]' \
+  '--no-auto-config[disable all config files except those next to executable]' \
+  '(-h --help)'{-h,--help}'[display help information]' \
+  '--help-all[display help including generic Qt options]' \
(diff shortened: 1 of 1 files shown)


Completion/X/Command/_sioyek
```diff
@@ -0,0 +1,32 @@
+#compdef sioyek
+
+local context state line
+typeset -A opt_args
```
okapia: These local variables are only needed if you've used the states feature of `_arguments`. As you've not used that feature, these lines aren't needed.

Completion/X/Command/_sioyek
```diff
@@ -0,0 +1,32 @@
...
+
+_arguments -S \
+    '--new-instance[create a new instance of sioyek when opening a file]' \
```
okapia: Our convention is only two spaces for indentation in completion functions. With all the descriptions, lines tend to get long so less indentation helps.

Completion/X/Command/_sioyek
```diff
@@ -0,0 +1,32 @@
...
+    '--instance-name[select a specific sioyek instance by name]:name: ' \
```
okapia: I'm not familiar with sioyek but if the option allows `=` between the option and it's argument, add the `=`. This indicates that it *must* be `--instance-name name`.

Completion/X/Command/_sioyek
```diff
@@ -0,0 +1,32 @@
...
+    '--xloc[set x position within page]:x-coordinate: ' \
+    '--yloc[set y position within page]:y-coordinate: ' \
+    '--window-id[apply command to specific window id]:window id: ' \
+    '--shared-database-path[path for shared bookmarks/highlights]:database path:_files' \
+    '--verbose[print extra information]' \
+    '--wait-for-response[wait for command to finish before returning]' \
+    '--no-auto-config[disable all config files except those next to executable]' \
+    '(-h --help)'{-h,--help}'[display help information]' \
+    '--help-all[display help including generic Qt options]' \
+    '*:pdf file:_files -g "*.pdf(.)"'
```
okapia: The normal idiom is to have `(-.)` as the glob qualifier. So either normal files or symbolic links to normal files.

Completion/X/Command/_sioyek
```diff
@@ -0,0 +1,32 @@
+#compdef sioyek
```
okapia, line 1: Thanks for the contribution.

It can also be worth trying to persuade the upstream (sioyek) project to carry the completion because they are more likely to keep the options up-to-date. If not, we're also happy to include the function.

Completion/X/Command/_sioyek
gi1242: It doesn't allow an = (I just tested it).

Completion/X/Command/_sioyek
gi1242: Thanks for going through this so quickly and your careful comments. I just implemented all your suggestions and will force push in a second.

I'll also send it upstream. If they take it, I'll post back and remove it from here, so we don't have two diverging branches.

okapia:
Merged. Thanks for the contribution.

And let us know if you manage to persuade upstream to carry the file.

gi1242:
> And let us know if you manage to persuade upstream to carry the file.

Filed an issue here: 

https://github.com/ahrm/sioyek/issues/1615

I'll post back if they plan on shipping it.

okdana:
jsyk: since this was accepted up-stream i've removed it from master, per discussion with oliver. but i'll keep it in 5.9 for now. hopefully packagers will have accounted for it by the time 5.10 is released

gi1242:
> jsyk: since this was accepted up-stream i've removed it from master, per discussion with oliver. but i'll keep it in 5.9 for now. hopefully packagers will have accounted for it by the time 5.10 is released

Thanks; I was waiting for it to be included in my distro before posting back... but I think that's gonna be a while. Removing it for 5.10 is a reasonable.

PR #174, opened 2026-06-20
completion: _gcc: coroutines-ts was merged in 2020; use the final flag

hhoffstaette:
Noticed that `_gcc` was still trying to complete `-fcoroutines-ts`, which hasn't been a thing since 2020. Update to `-fcoroutines` and move the definition to gcc so that it works for both.


okdana:
merged as 9a441838, ty

